PR #94, opened 2024-10-21
Suggest valid alternatives to invalid citation keys (#88)

### LINKED ISSUE #88: Give alias reminder when supplied id contains disallowed characters
Result running current version:
```
% autobib get 'doi:10.1016/0021-8693(85)90145-0'
ERROR Invalid bibtex entry key: doi:10.1016/0021-8693(85)90145-0
  Suggested fix: use an alias which does not contain disallowed characters: {}(),=\#%"
% autobib alias add aschbacher_scott_1985 'doi:10.1016/0021-8693(85)90145-0'
% autobib get 'doi:10.1016/0021-8693(85)90145-0' 
ERROR Invalid bibtex entry key: doi:10.1016/0021-8693(85)90145-0
  Suggested fix: use an alias which does not contain disallowed characters: {}(),=\#%"
```
In the third run above, it would be nice to remind the user that `aschbacher_scott_1985` is an alias. This should be easy to implement once #86 is merged.

wupr:
Fixes #88.

Also does some renaming.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -43,14 +43,12 @@ use self::{
         CitationKey, EntryData, RawRecordData, RecordData, RecordDatabase, RowData,
     },
     logger::Logger,
-    record::{
-        get_remote_record, GetRemoteRecordResponse, Record, RecordResponse, RecordRowResponse,
-    },
+    record::{get_remote_record, GetRemoteRecordResponse, Record, RecordRowResponse},
 };
 pub use self::{
     entry::Entry,
     http::HttpClient,
-    record::{get_record, get_record_row, Alias, RecordId, RemoteId},
+    record::{get_record_row, Alias, RecordId, RemoteId},
 };
 
 #[derive(Parser)]
@@ -299,8 +297,7 @@ fn run_cli(cli: Cli) -> Result<()> {
         Command::Alias { alias_command } => match alias_command {
             AliasCommand::Add { alias, target } => {
                 info!("Creating alias '{alias}' for '{target}'");
-                let row = record_db.initialize_row(&target)?;
-                match get_record_row(row, target, &client)? {
+                match get_record_row(&mut record_db, target, &client)? {
                     RecordRowResponse::Exists(_, row) => {
                         if !row.apply(row::add_alias(&alias))? {
                             error!("Alias already exists: '{alias}'");
@@ -371,8 +368,7 @@ fn run_cli(cli: Cli) -> Result<()> {
             }
         }
         Command::Edit { citation_key } => {
-            let row = record_db.initialize_row(&citation_key)?;
-            match get_record_row(row, citation_key, &client)? {
+            match get_record_row(&mut record_db, citation_key, &client)? {
                 record::RecordRowResponse::Exists(record, row) => {
                     edit_record_and_update(row, record)?;
                 }
@@ -402,7 +398,7 @@ fn run_cli(cli: Cli) -> Result<()> {
             ignore_null,
         } => {
             // Collect all entries which are not null
-            let valid_entries = validate_and_retrieve_entries(
+            let valid_entries = retrieve_and_validate_entries(
                 citation_keys.drain(..),
                 &mut record_db,
                 &client,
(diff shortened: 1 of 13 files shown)


src/main.rs
```diff
@@ -658,99 +661,142 @@ fn choose_canonical_id(
...
-                        error!("Null record: {remote_id}");
-                    }
-                    None
+    let valid_entries = citation_keys.filter_map(|citation_key| {
+        match get_record(record_db, citation_key, client) {
+            Err(err) => {
+                error!("{err}");
+                None
+            }
+            Ok(RecordResponse::Exists(record)) => get_bibtex_entry_from_record(record_db, record),
```
alexrutar, line 718: Should this be wrapped inside a transaction? The call to the `linked keys` can be done inside the same transaction. Then the `get_valid_keys` can be implemented as a function on a `RecordRow` directly.

alexrutar (commented):
Looks good to me! Only the comment about wrapping some operations inside a transaction.

alexrutar:
Also, it would be good to add some tests which verify this functionality. The `doi:` example with brackets is a nice one to keep in mind.

alexrutar:
Oh I noticed one more weird thing: if you mistakenly run (like I did)
```
cargo run -- alias add 'doi:10.1016/0021-8693(85)90145-0' alias
```
then the error message is weirdly formatted? Where is the `error:` coming from?

src/main.rs
wupr: There are several ways to merge the transactions. Which one do you have on your mind?

1. `RecordResponse::Exists` can carry a `Vec<String>` of alternative keys as well as the `Record`. But the alternatives are almost never needed.
2. `RecordResponse` can have another variant `InvalidKey(Vec<String>)`. This would mean moving the BibTeX validation logic into `record.rs`, which is pretty bad.
3. `RecordResponse::Exists` can forward the row transaction. This could have lifetime issues.

alexrutar: I mean to use the `get_record_entry` function here, and match on the `RecordEntryResponse` (which passes the row back); and then either commit it or pass it down to the `get_bibtex_entry_from_record`.

wupr: I see. So getting rid of `get_record` and `RecordResponse` completely then?

alexrutar: Yeah exactly, I think this would work the best.

alexrutar: And the `get_valid_keys` should have the same signature as, e.g., the `get_referencing_keys` function.

alexrutar:
Maybe the `doi:` example with brackets could also be added to the `README.md` as well (I guess better to change this in #95)

src/main.rs
alexrutar: Also in the `get_valid_keys` impl you can use `Vec::retain` to modify in-place (instead of filter / iter): https://doc.rust-lang.org/std/vec/struct.Vec.html#method.retain

src/main.rs
```diff
@@ -784,19 +797,8 @@ fn get_bibtex_entry_from_record(
...
-            .apply(row::get_referencing_keys)?
-            .into_iter()
-            .filter(|k| is_entry_key(k))
-            .collect();
-        row.commit()?;
-        return Ok(alternative_keys);
-    }
-    Ok(vec![])
+fn get_valid_keys(row: &RecordRow) -> Result<Vec<String>, rusqlite::Error> {
+    let mut alternative_keys = row.apply(row::get_referencing_keys)?;
```
alexrutar, line 801: Maybe `referencing_keys` instead of `alternative_keys`? Also maybe a better function name is `get_valid_referencing_keys`?

alexrutar (commented):
I realize `main.rs` is really quite large and should be split up at some point. But doing that in this PR isn't worth it.

I think I agree with removing the `get_record` function, and also the signature changes for `get_record_row`. However, I suppose it could happen in the future that we will need these versions again, but I guess it would be easy to re-implement if necessary.

alexrutar:
Looks like the invalid `()` is causing errors in the mitmproxy generation step! This was a bit of an oversight.

alexrutar:
You can pull the test commit out into a separate PR, where we also try to think of a more robust solution for the mitmproxy flows generation. I am happy for this PR to be committed as is, it looks great! Thanks for implementing it.

wupr:
Great! Thanks for the review. Splitting and merging now.

PR #97, opened 2024-10-21
Add test for BibTeX key validation

wupr:


src/main.rs
```diff
@@ -128,6 +135,10 @@ enum Command {
         /// Write output to file.
         #[arg(short, long)]
         out: Option<PathBuf>,
+        /// The output format.
+        /// Select `null` to suppress BibTeX output and disable checks on citation keys.
```
alexrutar, line 139: I think you need to leave a space after the first line, or it creates an overflowed help message when you run `autobib help get`.

src/main.rs
```diff
@@ -538,10 +558,13 @@ fn run_cli(cli: Cli) -> Result<()> {
...
                 &mut record_db,
                 &client,
+                format,
                 ignore_null,
             );
 
-            output_entries(out.as_ref(), valid_entries)?;
+            if format != OutputFormat::Null {
+                output_entries(out.as_ref(), valid_entries)?;
+            }
```
alexrutar, line 567: I would probably prefer if the `write_entries` / `output_entries` signature were modified to include the `OutputFormat` as an argument, to make it easier to implement new formats in the future.

src/main.rs
```diff
@@ -730,19 +754,26 @@ fn retrieve_and_validate_single_entry(
...
+                } = record;
+                let entry = validate_bibtex_entry(key, data, &row).map(|entry| (entry, canonical));
+                row.commit()?;
+                Ok(entry)
+            }
+            OutputFormat::Null => {
+                row.commit()?;
+                Ok(None)
+            }
+        },
```
alexrutar, line 776: I don't really think this optimization is worthwhile, it increases the complexity of this implementation for a rather uncommon function call.

wupr: What would you suggest we do? `validate_bibtex_entry` just handles the key validation, so I think it makes sense to skip it altogether.

alexrutar: Oh I see, of course when the `Entry` is generated, the error is already cause.

src/entry.rs
```diff
@@ -9,37 +9,55 @@ use serde_bibtex::{de::Deserializer, to_string_unchecked, validate::is_entry_key
...
-    pub key: String,
-    pub record_data: D,
+pub struct BibtexKey {
+    key: String,
 }
 
-impl<D: EntryData> Entry<D> {
+impl BibtexKey {
     /// Create a new entry data with the provided key.
     ///
```
alexrutar, line 22: Typo: should be updated

src/entry.rs
```diff
@@ -9,37 +9,55 @@ use serde_bibtex::{de::Deserializer, to_string_unchecked, validate::is_entry_key
...
-    ///
-    /// # Safety
-    /// The caller is required to guarantee that the key does not contain any characters which are
-    /// invalid BibTeX entry key characters, as accepted by the [`serde_bibtex::validate::is_entry_key`] method.
-    pub(crate) fn new_unchecked(key: String, record_data: D) -> Self {
-        Self { key, record_data }
+/// A single regular entry in a BibTeX bibliography.
+#[derive(Debug, PartialEq)]
+pub struct Entry<D: EntryData> {
+    pub key: String,
```
alexrutar, line 50: Should this rather be a `BibtexKey`?

alexrutar: If this is changed, I guess the `Serializer` impl will break (but is easy to fix), but hopefully nothing else.

alexrutar: Also, if this is changed, for convenience, we should implement `From<BibtexKey> for String`

alexrutar: If we go down this route, it should also implement `AsRef<str>` which is useful to fix the errors in `main.rs`. In which case I would even `BibtexKey` being a "newtype"-style struct.

alexrutar: Oh I just realized I literally implemented this once already: https://docs.rs/serde_bibtex/latest/serde_bibtex/token/struct.EntryKey.html

But the API is quite bad, if we want to merge these changes in, then I should update upstream...

wupr: I've added the suggested implementations. Happy to replace with `EntryKey` if you want to update upstream.

alexrutar:
Let me fix upstream and see how it goes.

alexrutar:
I'm most of the way through implementing this upstream; unfortunately, a pre-requisite was to fix all of the horrifying error handling in that crate which will take a bit of time.

Issue #96, opened 2024-10-21
Clap errors inconsistent in style with our errors


wupr:
_Originally posted by @alexrutar in https://github.com/autobib/autobib/issues/94#issuecomment-2426577872_:

```
cargo run -- alias add 'doi:10.1016/0021-8693(85)90145-0' alias
```
[Clap] is processing the errors up-front (during argument parsing) and immediately complaining about the function signature. It's just a bit confusing that it is formatted as `error:` I guess the best way to fix this is (1) figure out how to reformat clap error message formatting; and then (2) not print the citation key twice.

alexrutar:
From: https://github.com/autobib/autobib/pull/98#discussion_r1813815642

We can also make our message headings boldface and lower case. But we still need to deal with clap inserting an extra `:` (or do that ourselves, too).

PR #98, opened 2024-10-22
Improve database state management

### LINKED ISSUE #91: `autobib update` improvements
- [ ] Add some sort of update-mode option to update, to either: (1) prefer local, (2) prompt for changes from remote, or (3) prefer remote
- [ ] Add `--from` option for update, to read the 'update' data from a local bibtex file

alexrutar:
The primary goal of this PR is to improve the internal database state management. The main change is the generalization of the previous `Row` implementation. The earlier implementation of `Missing` is a somewhat leaky abstraction, since `Missing` could in fact correspond to two distinct states: missing, but in the `NullRecords` table; and missing, and not in the database at all. This PR uses three individual different states which can be associated with citation key.

1. The citation key corresponds to a row in the `Records` table. This is the same `RecordRow` struct from before.
2. The citation key corresponds to a row in the `NullRecords` table. This is the new `NullRecordRow` struct.
3. The citation key is a remote id, and does not correspond to anything in the database. This is the new `Undefined` struct (better naming?).

In order to initialize the state, the `RecordDatabase` offers two new functions. The first is `state_from_record_id`, which given a `RecordId`, determines the current state of this `RecordId` as can be determined by the database. There are exactly 5 non-overlapping possibilities, which correspond to the variants of the `db::row::State` enum:

- The row exists.
- The record id is a remote id, and it is null.
- The record id is a remote id, and is undefined in the database.
- The record id is a null alias.
- The record id is a remote id, and is invalid.

There is also an option when we know the `RecordId` is not an alias (this is the `state_from_remote_id` function, and is quite similar).

There are also conversion methods between states.

- `RecordRow::delete` results in an `Undefined` (since it is removed from the database).
- `NullRecordRow::delete` results in an `Undefined` (since the null is removed from the database).
- `Undefined::insert` results in a `RecordRow`.
- `Undefined::set_null` results in a `NullRecordRow`.
- `NullRecordRow::insert` deletes the row in the `NullRecords` table and creates data in `Records` table, resulting in a `RecordRow`.

As an indicator that the state management is better, just while implementing this PR, I found 2 implementation errors and 3 unintended (but probably rare) oversights. I think this is a good sign that the abstractions are better, since the resulting code is in my opinion much more robust!

This PR will also resolve #91 when it is done.

TODOS:
- [x] Better naming? Rename `row` module to `state`; better name for `Undefined`?
- [x] Actually implement the `UpdateMode` changes.
- [x] Fix the update logic bug.


src/db/row.rs
```diff
@@ -1,58 +1,136 @@
...
+/// A representation of the database state beginning with an arbitrary [`RemoteId`].
+#[derive(Debug)]
+pub enum RemoteState<'conn> {
+    /// The `Records` row exists.
+    Exists(RemoteId, RecordRow<'conn>),
+    /// The `Records` row does not exist and the `NullRecords` row exists.
+    NullRemote(RemoteId, NullRecordRow<'conn>),
+    /// The `Records` and `NullRecords` rows do not exist.
+    UndefinedRemote(RemoteId, Undefined<'conn>),
+    /// The remote id is invalid.
```
alexrutar: Maybe it makes more sense to not pass the `RemoteId` back here (and instead just get a reference up-front)? But I wrote it in this way to be consistent with the `State` implementation, where ownership is necessary for the `resolve` step.

src/db/row.rs
```diff
@@ -1,58 +1,136 @@
...
+    }
 }
 
-/// A representation of a row in the `Records` table which is missing.
+/// The database state when:
+/// 1. There is no row in `Records`.
+/// 2. There is no row in `NullRecords`.
 #[derive(Debug)]
-pub struct MissingRecordRow<'conn> {
+pub struct Undefined<'conn> {
```
alexrutar: Better name for `Undefined`?

src/error.rs
```diff
@@ -28,6 +26,10 @@ pub enum Error {
     DatabaseError(#[from] DatabaseError),
     #[error("Provider error: {0}")]
     ProviderError(#[from] ProviderError),
+    #[error("Reference source returned a key corresponding to a null record: {0}")]
+    UnexpectedNullRemoteFromProvider(String),
+    #[error("Reference source returned a remote id which is invalid: {0}")]
+    InvalidRemoteIdFromProvider(RecordError),
```
alexrutar, line 32: Should these error variants be merged into the `ProviderError`?

src/main.rs
```diff
@@ -463,25 +483,64 @@ fn run_cli(cli: Cli) -> Result<()> {
...
+                    println!("Undefined alias: {alias}");
+                }
+                InfoReportType::Canonical => {
+                    bail!("No canonical id for undefined alias '{alias}'");
+                }
+                InfoReportType::Valid => {
+                    if !is_entry_key(alias.name()) {
+                        error!("Invalid BibTeX: {}", alias.name());
+                    }
+                }
```
alexrutar, line 520: Is it reasonable to check validity even when the alias is not actually in the database? I can easily imagine this use-case; just wondering if this is a sensible return type.

src/provider.rs
```diff
@@ -60,6 +60,33 @@ pub(crate) fn lookup_validator(provider: &str) -> Option<Validator> {
...
+            Some(data) => Ok(ProviderOutcome::Data(data)),
+            None => Ok(ProviderOutcome::Null),
+        },
+        Either::Right(referrer) => match referrer(remote_id.sub_id(), client)? {
+            Some(new_remote_id) => Ok(ProviderOutcome::Reference(new_remote_id)),
+            None => Ok(ProviderOutcome::Null),
+        },
+    }
+}
+
```
alexrutar, line 89: Should we actually just expose an entirely different API from the `provider` module? I.e., also do not expose the `lookup_validator` function, and just return another function like `validate_remote_id(&RemoteId) -> bool`. Then the `lookup_validator` and `lookup_provider` are just internal lookup tables for implementation convenience.

alexrutar (commented):
Just pointing out some things which I was not sure about while implementing.

src/db/row.rs
wupr: Not passing it back might make it clearer that the id won't see any change. I feel it's OK either way.

src/db/row.rs
wupr: This is a hard one. `Idle`?

src/error.rs
wupr: I think they should.

src/main.rs
wupr: This might be misleading (by not telling the user that the alias doesn't exist). I think `autobib info <alias> valid` is a bad way to check whether an alias is a valid citation key.

Also, we should make the info report type something like a bit flag.

src/provider.rs
wupr: Sounds reasonable. I don't have any objection.

src/logger.rs
```diff
@@ -10,6 +10,18 @@ use std::{
...
 
+macro_rules! suggest {
+    () => {
+        eprintln!()
+    };
+    ($($arg:tt)*) => {
+        eprint!("{}", ::crossterm::style::Stylize::blue("HINT "));
+        eprintln!($($arg)*);
+    };
+}
```
wupr, line 21: Should we make our message headings lowercase and boldface?

wupr:
How about refactoring `state.rs` and `state/operations.rs` so that each state-transaction wrapper (`RecordRow`, `NullRecordRow`, `Undefined`) is in its own file with the applicable operations? E.g. `RecordRow` and its `impl` block would share a file with `get_null_attempted` and `get_null_row_data`.

wupr:
I feel `ProviderOutcome` and `RemoteRecordResponse` should have similar names; in my head the latter is the result of recursing over the former. I would even call them `RemoteResponse` and `RecursiveRemoteResponse` respectively. `get_provider_outcome` and `get_remote_record` would then also be renamed accordingly. And `get_record_row_remote_resolve` is just recursing over both `RemoteResponse` and `RemoteIdState`.

wupr (commented):
Trying to finish the review but GitHub requires me to leave a comment, like this one.

src/main.rs
```diff
@@ -105,6 +105,9 @@ enum Command {
         /// Delete without prompting.
         #[arg(short, long)]
         force: bool,
+        /// Delete a row in the usual records table or the null records table.
+        #[arg(short, long)]
+        null: bool,
```
wupr, line 110: Is this meant to say "Delete a row in the null records table"?

But actually, I think it might be useful to handle multiple citation keys here (in a future PR), so I would instead do a `--delete-null` option, with enables deleting null records as well as usual records.

src/main.rs
```diff
@@ -490,18 +549,26 @@ fn run_cli(cli: Cli) -> Result<()> {
...
-                    let row = missing.insert_and_ref(&data, &remote_id)?;
-                    (row, data)
+                    let row = undefined.insert_and_ref(&data, &remote_id)?;
+                    (row, data, remote_id)
+                }
+                RemoteIdState::Null(_, _) => {
+                    unreachable!(
+                        "A local record should not be present in the `NullRecords` table."
+                    );
+                }
```
wupr, line 559: This is not logically unreachable code (as the database could be externally modified), so `unreachable!` shouldn't be used here. Maybe `bail!`?

src/main.rs
```diff
@@ -490,18 +549,26 @@ fn run_cli(cli: Cli) -> Result<()> {
...
+                RemoteIdState::Invalid(_) => {
+                    unreachable!("All identifiers for local records should be valid.");
                 }
```
wupr, line 562: I was going to check whether this is logically unreachable, but noticed that `RemoteIdState::Invalid` is never created (I could only find two consumptions, including this one).

src/main.rs
alexrutar: I don't know if it's possible for this to be a bit flag; what does it mean to ask for 'canonical' and 'valid' at the same time? I guess there is the "report" itself, and then the methods to access various parts.

src/db/row.rs
alexrutar: I've changed it to not pass the ID back, I think it is a bit more understandable this way.

src/logger.rs
alexrutar: Sure, that sounds reasonable to me! Maybe I will save this for a different PR (when fixing #96, where we need to deal with a few other things).

src/main.rs
alexrutar: Good catch, this error category is only created when you call the `resolve` method on a `RecordId` and not when already passing a `RemoteId`

src/provider.rs
alexrutar: Something I just realized is that the `validate_remote_id` cannot have a call signature which accepts a `&RemoteId`, since by definition once it is a `RemoteId` it is already valid!

But I guess we could just as easily write `validate_remote_id(provider: &str, sub_id: &str) -> bool`. I will implement this.

src/main.rs
alexrutar: Should we assume that the database could be externally modified in the `db.rs` code as well? If e.g. the binary data is invalid, the code will also panic.

src/db/row.rs
alexrutar: Actually, maybe `MissingRow` is quite precise here?

src/main.rs
wupr: Yeah, maybe we should handle data corruption more gracefully.

I think that it's not that bad to `panic!` in the above; just that `unreachable!` is semantically inaccurate.

alexrutar:
Most recent commit is in progress; just pushing to avoid more conflicts. Will continue tomorrow (or you can pick it up and continue if you wish)

src/main.rs
alexrutar: Makes sense; well, at least in this spot, recovering is quite sensible. In the binary data, even returning an error feels dangerous because if the database is malformed we pretty much want to bail as fast as possible and ask the user to try to fix it / figure out what is wrong using the `autobib util check` command. Because pretty much all of the database operations are undefined once the database is in an undefined state.

alexrutar:
TODOs:
- [ ] Remote the `default` from` default_value_t` (since it uses `default` automatically)
- [ ] Make the `report` argument into a flag in `autobib info`
- [ ] Fill in `todo!`s in `autobib update`:
  - [ ] Implement a more general merge strategy for the `RecordData`.

For the more general merge strategy, I think it would be best for the `try_merge` to accept an additional 'conflict resolution' callback of the form
```rust
FnMut(&key, our_value, their_value) -> inserted_value`
```
I think this callback needs to be infallible, or the merge could end up in an undefined state; errors such as an io error during prompt read should just log the error and then return `our_value` or whatever.

Then in the `autobib update` function, define the appropriate callbacks.

wupr:
- This PR is already at +984, -535 lines. Let's do the `autobib info` change in a new PR.

Issue #101, opened 2024-10-23
Set state via environment variables


alexrutar:
We should allow setting some state through environment variables.

The main thing I could think of was something like `AUTOBIB_DATABASE` to set the default database location.

Maybe there are some other sensible choices?

Issue #62, opened 2024-09-28
Support sourcing keys from `.aux` file


alexrutar:
The `.aux` file should also contain relevant keys... maybe this will be useful, the format should be quite simple, we could also support that.

PR #108, opened 2024-10-24
Housekeeping

### LINKED ISSUE #107: Remove some dependencies
I think we should remove the `either::Either` dependencies (since I constantly get confused about what `Left` and `Right` mean) and the the `lazy_static` dependency since it is now in the stdlib: https://github.com/rust-lang-nursery/lazy-static.rs/issues/214.

Also take this as a chance to see if there is anything else we should remove.

alexrutar:
Fixes #107 #101 #96 #62.

These changes are basically uncontroversial, and were quite simple to implement. Mostly just clearing out some of the issues queue.

Most notably / for discussion, I've deleted the field count limit in the binary data format. Since this is fully backwards compatible (i.e. every binary data format which was valid before is still valid), this does not require bumping the binary format version.

My reasoning for this is as follows.

1. It makes the implementation a bit simpler on our end. Most notably, since there is no length restriction, it opens the future option to implement types like `struct DataKey(String)` and `struct DataKey(Value)`, and then have an infallible insertion function. For instance, this would allow better error messages within a user-provided struct, since inside the prompt loop we can try to construct the relevant types, and prompt to update if it fails.
2. If the user decides to insert X gigabytes into their SQLite database and things get slow, I feel like that's not really our problem. Since there is already a (relatively strict) data length limit per key; I think it's 65KB, they would really have to try very hard to hit database sizes which cause problems.
3. There is less logic to maintain on our end.


alexrutar:
@wupr What do you think of the `AUTOBIB_DATABASE_FILE` environment variable naming?

wupr:
> @wupr What do you think of the `AUTOBIB_DATABASE_FILE` environment variable naming?

I think `PATH` is more accurate than `FILE`. `AUTOBIB_DATABASE_PATH` or `AUTOBIB_DB_PATH`?

alexrutar:
`AUTOBIB_DATABASE_PATH` makes sense to me! I think slightly more verbose is better.

PR #109, opened 2024-10-25
Improve `autobib local`

### LINKED ISSUE #89: CLI improvements
Currently, starting with no `local:wu_2024` in the database:
```
% autobib local wu_2024                   
% autobib edit wu_2024
ERROR Cannot edit null record 'wu_2024'
% autobib local wu_2024
```

Several possible improvements here:

- [x] ~~Set the default verbosity level to `INFO` and add some more messages?~~ E.g. `autobib local` could say that `local:wu_2024` has been created (or it already exists), as a reminder that `wu_2024` is not the full id.
- [x] Automatically create an alias `wu_2024` with `autobib local wu_2024`, with an optional `--no-alias` flag?
- [x] `autobib local` should open the editor if neither `--from <FILE>` nor `--no-edit` is supplied.
- [x] `autobib edit` should say "Cannot edit non-existent record" as "null record" has a specific meaning.

wupr:
Fixes #89.

### DIFF
--- a/src/error/record.rs
+++ b/src/error/record.rs
@@ -15,6 +15,7 @@ pub enum RecordErrorKind {
     InvalidProvider,
     InvalidSubId,
     RecordIdIsNotAlias,
+    RecordIdIsNotRemoteId,
     EmptyAlias,
 }
 
@@ -22,31 +23,44 @@ impl fmt::Display for RecordError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         write!(f, "Invalid key '{}': ", self.input)?;
         match self.kind {
-            RecordErrorKind::EmptyProvider => f.write_str("provider must be non-empty"),
-            RecordErrorKind::EmptySubId => f.write_str("sub-id must be non-empty"),
-            RecordErrorKind::EmptyAlias => f.write_str("alias must be non-empty"),
+            RecordErrorKind::EmptyProvider => {
+                f.write_str("provider must contain non-whitespace characters")
+            }
+            RecordErrorKind::EmptySubId => {
+                f.write_str("sub-id must contain non-whitespace characters")
+            }
+            RecordErrorKind::EmptyAlias => {
+                f.write_str("alias must contain non-whitespace characters")
+            }
             RecordErrorKind::InvalidProvider => f.write_str("provider is invalid"),
             RecordErrorKind::InvalidSubId => {
                 f.write_str("sub-id is invalid for the given provider")
             }
             RecordErrorKind::RecordIdIsNotAlias => f.write_str("alias must not contain a colon"),
+            RecordErrorKind::RecordIdIsNotRemoteId => f.write_str("remote id must contain a colon"),
         }
     }
 }
 
 #[derive(Error, Debug)]
 pub enum AliasConversionError {
-    #[error("alias must not contain a colon")]
-    IsRemoteId,
-    #[error("alias must be non-empty")]
-    Empty,
+    #[error("Invalid alias '{0}': alias must not contain a colon")]
+    IsRemoteId(String),
+    #[error("Invalid alias '{0}': alias must contain non-whitespace characters")]
+    Empty(String),
 }
 
 impl From<AliasConversionError> for RecordError {
     fn from(err: AliasConversionError) -> Self {
         match err {
-            AliasConversionError::IsRemoteId => todo!(),
-            AliasConversionError::Empty => todo!(),
+            AliasConversionError::IsRemoteId(input) => RecordError {
+                input,
+                kind: RecordErrorKind::RecordIdIsNotAlias,
+            },
+            AliasConversionError::Empty(input) => RecordError {
+                input,
+                kind: RecordErrorKind::EmptyAlias,
+            },
         }
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -172,12 +172,15 @@ enum Command {
     Local {
         /// The name for the record.
         id: String,
-        /// Edit record.
-        #[arg(long, action)]
-        edit: bool,
         /// Create local record from bibtex file.
         #[arg(short, long, value_name = "PATH")]
         from: Option<PathBuf>,
+        /// Do not create the alias `<ID>` for `local:<ID>`.
+        #[arg(long)]
+        no_alias: bool,
+        /// Save the local record without editing.
+        #[arg(long)]
+        no_edit: bool,
     },
     /// Generate records by searching for citation keys inside files.
     ///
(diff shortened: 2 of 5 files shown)


src/main.rs
```diff
@@ -563,18 +572,33 @@ fn run_cli(cli: Cli) -> Result<()> {
                 }
             };
 
-            if edit {
+            if !no_alias {
+                match Alias::from_str(&id) {
+                    Ok(alias) => {
+                        info!("Creating alias '{alias}' for '{canonical}'");
+                        row.apply(state::add_alias(&alias))?;
```
alexrutar, line 579: The `add_alias` operation is fallible (if the alias exists, it returns `false`).

1. Should the return type of `add_alias` be unpacked into an actual enum, for clarity?
2. We should definitely do something if alias creation fails.

src/main.rs
```diff
@@ -563,18 +572,33 @@ fn run_cli(cli: Cli) -> Result<()> {
...
+                    }
+                    Err(error::AliasConversionError::Empty) => {
+                        warn!("Alias not created for '{canonical}'.");
+                    }
```
alexrutar, line 583: It is weird that this could happen. We should be blocking the creation of a `local:` record id if the chosen key is empty, I guess?

alexrutar (changes requested):

alexrutar:
I am experimenting with the 'request changes' flow, no idea how it actually works!

src/main.rs
wupr: I overlooked this. Will push a commit to fix this now.

I tried tagging `add_alias` with `#[must_use]` but it doesn't work because the return type is `Result`, which is used (and in fact, all `Result`s are must-use). I don't know if we can require the `Ok` variant to be used.

I think it's fine to leave it as `bool`.

alexrutar: If you create a custom return type and annotate it with `#[must_use]`, what happens?

alexrutar: Does the `#[must_use]` "reach downwards"? Or is it just forgotten immediately when unwrapping the result?

src/main.rs
wupr: I suppose this was actually a bug in the `RemoteId::local` function. Running `cargo run -- local ''` would have been successful before.

Fixed now, although I'm not completely happy with the `unreachable!` here. I wonder if there is a more satifying solution.

src/record/key.rs
```diff
@@ -196,13 +196,19 @@ impl RemoteId {
     }
 
     /// Create a new `local` [`RecordId`].
-    pub fn local(sub_id: &str) -> Self {
+    pub fn local(sub_id: &str) -> Result<Self, RecordError> {
         let mut full_id = String::with_capacity(9);
```
alexrutar, line 200: Haha the 9 is a bit random (my fault), maybe a nice round number like `16` is more sensible? I guess we need at least 6 bytes for the `local:`, and then I feel like 10 bytes is a good amount of space for most cases to avoid realloc.

wupr: Nice round number like 0b10000! Done.

alexrutar: Feels like unnecessary optimization anyway, mainly I got slightly bothered by the 9.

alexrutar (approved):

src/record/key.rs
alexrutar: Great thanks!

src/main.rs
alexrutar: Yeah fair enough, it looks good to me! I agree that the `unreachable` is a bit unsatisfying.

A potentially less arbitrary solution would be to require that `local:` keys always satisfy the alias rules.

alexrutar: I.e. change the `local` initialization code to be infallible, but require passing an `Alias`.

src/main.rs
wupr: I think that's the problem. Unwrapping or propogating seems enough to satify the compiler. Creating a custom return type for `add_refs_impl` and annotating it with `#[must_use]` works though; it creates 5 warnings at other lines.

src/main.rs
wupr: Right, this would make sense if we also insist that the `Alias` cannot exist already.

alexrutar:
By the way, for backwards compatibility, we should probably not modify the `is_valid_id` in the `local` provider code since this could randomly break things if someone is using the database. Instead I would advocate for inserting a warning there (that this has been changed, and a suggested command to fix it) and then we can make it an error in the future.

wupr:
That's a good point. But I feel the probability that someone has run `autobib local '' --edit` and put meaningful data in it is extremely low, and they wouldn't have been able to do anything about it after that (e.g. `autobib get local:` wouldn't work). It should suffice to add this change and a suggested fix to the release notes.

wupr:
Actually... how would they fix it? `autobib delete local:` won't work, either!

alexrutar:
In the `get record` flow, we never pass through a `RemoteId` at all: as far as I am aware, we just check validity using the `is_valid_id` step. My opinion is basically to implement everything as discussed before - including requiring that a local remote id be constructed from an `Alias`, for instance; but then inside the `is_valid_id` function just include a warning.

alexrutar:
And then deprecating the support for invalid IDs locally is just a change in a single place (i.e. `provider::local::is_valid_id`)

wupr:
On the current main branch, in `impl TryFrom<RecordId> for RemoteId`, the non-emptiness of the sub-id is checked. But this check is not in `RemoteId::new`, which is used by the `TryFrom` and `RemoteId::from_parts`. This means that `from_parts` misses the check. So in the last commit I moved the non-emptiness check from `TryFrom` to `new` and I think this helps prevent future bugs.

On the other hand, `RemoteId::local` uses `RemoteId::new_unchecked` so it bypasses all checks. What I ended up doing is add a non-emptiness check to `RemoteId::local`, without changing `is_valid_id` at all. And I've just realised that this does not break backward compatibility, because if you have `local:` in your database already, you can still call `autobib get local:` and `autobib delete local:`.

src/error/record.rs
```diff
@@ -38,17 +38,23 @@ impl fmt::Display for RecordError {
...
 #[derive(Error, Debug)]
 pub enum AliasConversionError {
-    #[error("alias must not contain a colon")]
-    IsRemoteId,
-    #[error("alias must be non-empty")]
-    Empty,
+    #[error("Invalid alias '{0}': alias must not contain a colon")]
+    IsRemoteId(String),
+    #[error("Invalid alias '{0}': alias must be non-empty")]
+    Empty(String),
```
alexrutar, line 44: The reason initially for not including this message initially to suppress the weird warning when `Alias` is read from the command line.

alexrutar: e.g. the error message which occurs writing something like `autobib alias add '::' <something>`

alexrutar: But overall, I think this handling of errors is better. Is there a way to special-case the CLI handling somehow?

wupr: We could defer the type conversion, i.e. accept any string and convert manually.

alexrutar: Sounds good to me!

PR #114, opened 2024-10-27
Various fixes for issues involving citation keys and aliases

alexrutar:
As discussed offline, there are some shortcomings with the current implementation:

- [x] Valid bibtex keys cannot contain whitespace.
- [x] A commit was dropped from a previous PR which had some fixes for alias checking

### DIFF
--- a/src/error/record.rs
+++ b/src/error/record.rs
@@ -2,6 +2,8 @@ use std::fmt;
 
 use thiserror::Error;
 
+use super::ShortError;
+
 #[derive(Error, Debug)]
 pub struct RecordError {
     pub input: String,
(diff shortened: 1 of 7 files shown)


wupr (changes requested):
The messages have been mixed up in the reordering.

src/error/record.rs
```diff
@@ -19,37 +21,43 @@ pub enum RecordErrorKind {
...
-            }
-            RecordErrorKind::RecordIdIsNotAlias => f.write_str("alias must not contain a colon"),
-            RecordErrorKind::RecordIdIsNotRemoteId => f.write_str("remote id must contain a colon"),
+            RecordErrorKind::EmptyProvider => "provider must contain non-whitespace characters",
+            RecordErrorKind::EmptySubId => "sub-id must contain non-whitespace characters",
+            RecordErrorKind::InvalidProvider => "alias must contain non-whitespace characters",
+            RecordErrorKind::InvalidSubId => "provider is invalid",
+            RecordErrorKind::RecordIdIsNotAlias => "sub-id is invalid for the given provider",
+            RecordErrorKind::RecordIdIsNotRemoteId => "alias must not contain a colon",
+            RecordErrorKind::EmptyAlias => "remote id must contain a colon",
```
wupr, line 33: The messages have been mixed up in the reordering.

wupr (changes requested):

src/error/record.rs
alexrutar: Fixed now! Good catch, I guess we should have tests to catch this, but we are missing tests for a lot of things more generally...

wupr (approved):
Thanks! Although the `match` block in `impl ShortError for RecordError` needs to be sorted again.

PR #117, opened 2024-10-27
Improve database consistency checks

alexrutar:
Changes:

- [x] Do not bail on first consistency error; instead, try to find as many as possible.
- [x] Attempt to fix consistency errors when possible.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -328,8 +328,22 @@ impl RecordDatabase {
                     .execute((canonical, key))?;
                 Ok(true)
             }
-            DatabaseFault::NullCitationKeys(count) => {
-                warn!("Deleting {count} citation keys which do not reference records");
+            DatabaseFault::NullCitationKeys(_) => {
+                let mut invalid_keys: Vec<String> = Vec::new();
+                {
+                    let mut stmt = tx.prepare(
+                        "SELECT name FROM CitationKeys WHERE record_key NOT IN (SELECT key FROM Records)",
+                    )?;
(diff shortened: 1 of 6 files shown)


src/db.rs
```diff
@@ -290,16 +287,60 @@ impl RecordDatabase {
 
     /// Validate the internal consistency of the database.
     ///
-    /// This does not modify the database.
-    pub fn validate(&mut self) -> Result<(), ValidationError> {
+    /// The `fix` callback is used to determine whether or not to apply a potentially destructive
+    /// database change. If the `fix` callback always returns false, then no modifications to the
+    /// database will be performed.
```
wupr, line 292: `fix` doesn't seem to be a callback argument

src/main.rs
```diff
@@ -760,9 +764,15 @@ fn run_cli(cli: Cli) -> Result<()> {
             RecordIdState::InvalidRemoteId(err) => bail!("{err}"),
         },
         Command::Util { util_command } => match util_command {
-            UtilCommand::Check => {
+            UtilCommand::Check { fix } => {
                 info!("Validating record binary data and consistency, and checking for dangling records.");
-                record_db.validate()?;
+                let faults = record_db.recover(fix)?;
+                if !faults.is_empty() {
+                    error!("Database contains critical errors!");
```
wupr, line 771: I think the message can be toned down a bit. Something like "Erroneous data found in the database."?

wupr (approved):

PR #118, opened 2024-10-28
Update README for v0.2.0 release

wupr:
- [x] `local:` validity changes
- [x] `autobib local` prompts for edit by default
- [x] `autobib delete` and `autobib update`


README.md
```diff
@@ -125,7 +125,7 @@ Then running `autobib get hochman-entropy` returns
 The record is identical to the record `zbl:1337.28015`, except that the citation key is the name of the alias.
 In order to distinguish from usual identifiers, an alias cannot contain the `:` colon symbol.
 
-Note that the characters `{}(),=\#%"` are not permitted in a BibTeX entry key.
+Note that the characters `{}(),=\#%"` and whitespaces are not permitted in a BibTeX entry key.
```
alexrutar, line 132: Maybe 'whitespace characters' is better? "Whitespaces" feels weird to me. Then we could link to https://doc.rust-lang.org/reference/whitespace.html

i.e. [whitespace characters](https://doc.rust-lang.org/reference/whitespace.html)

README.md
```diff
@@ -90,16 +91,19 @@ autobib source main.tex --out main.bib
...
 The data is stored locally in a [SQLite](https://www.sqlite.org/) database, which defaults to `~/.local/share/autobib/records.db`, so that subsequent runs are substantially faster.
-You can view and update the internally stored record with
+You can view and modify the internally stored record with
 ```bash
 autobib edit zbl:1337.28015
 ```
 If the record does not yet exist in your local record database, it will be retrieved before editing.
 
+You can also re-retrieve a record from the remote provider using the `autobib update` command, or remove one from the database using the `autobib delete` command.
+Run `autobib help update` and `autobib help delete` for more details.
```
alexrutar, line 105: 'for more detail'?

alexrutar (approved):
Thanks for the updates! Some minor comments but otherwise seems totally fine.

Looks like we are pretty close to a `v0.2.0` release.

README.md
wupr: How about "whitespace" and link to the reference? I think this reads better than two "characters" in the same phrase.

alexrutar: Works too!

README.md
wupr: I think "for more details" is fine. https://dictionary.cambridge.org/dictionary/english/detail

alexrutar: I guess I just prefer detail to details! But I agree both are grammatically OK.

wupr: I'll go with the majority opinion😬 https://books.google.com/ngrams/graph?content=for+more+detail%2Cfor+more+details&year_start=1800&year_end=2022&corpus=en&smoothing=3

PR #125, opened 2024-11-10
Implement custom wrapper around rusqlite transaction for additional logging

wupr:


alexrutar (approved):
Looks good to me!

src/db.rs
```diff
@@ -488,3 +490,51 @@ pub enum DeleteAliasResult {
...
+            where
+                F: FnOnce(&rusqlite::Row<'_>) -> rusqlite::Result<T>;
+            pub fn prepare(&self, sql: &str) -> rusqlite::Result<rusqlite::Statement<'_>>;
+            pub fn prepare_cached(&self, sql: &str) -> rusqlite::Result<rusqlite::CachedStatement<'_>>;
+        }
+    }
+}
+
+impl Drop for Transaction<'_> {
+    fn drop(&mut self) {
```
alexrutar, line 534: It just occurred to me that this should probably be `#[inline]`, since in a release build `debug!` should be stripped, which means this function is a no-op.

I mean, almost certainly, in a release build this function call will be entirely removed anyway (there's the joke that the answer to whether or not LLVM will inline a function call is 'yes'). But I guess it doesn't hurt anyway.

wupr: Done!

PR #126, opened 2024-11-10
Fix warning about existing alias for local record

wupr:


src/db/state/record.rs
```diff
@@ -106,6 +108,32 @@ impl<'conn> State<'conn, RecordRow> {
...
+        debug!(
+            "Ensuring alias '{alias}' refers to row_id '{}'",
+            self.row_id()
+        );
+        match get_row_id(&self.tx, alias)? {
+            Some(existing_row_id) => {
+                if existing_row_id == self.row_id() {
+                    return Ok(None);
+                } else {
+                    let RowData { canonical, .. } = get_row_data(&self.tx, existing_row_id)?;
```
alexrutar, line 125: It bothers me somewhat that the function `get_row_data` is implemented twice, but I can see why it is done this way here. I might be tempted to manually inline the `get_row_data` call here.

src/db.rs
```diff
@@ -101,6 +101,11 @@ fn get_row_id<K: CitationKey>(
         .optional()
 }
 
+fn get_row_data(tx: &Transaction, row_id: RowId) -> Result<RowData, rusqlite::Error> {
+    tx.prepare_cached(sql::get_record_data())?
+        .query_row([row_id], |row| RowData::try_from(row))
```
alexrutar, line 106: I think this can be `|row| row.try_into()`

alexrutar (commented):
I spotted a few other minor things while looking over this PR which I've also pushed now too.

alexrutar:
Just to clarify, what exactly does this PR change from the behaviour that was previously implemented?

wupr:
Before, running `autobib local foo` when `local:foo` and `foo` already exist (and pointing to the same record) would result in  the warning message "Alias 'foo' already exists. 'local:foo' will be a different record.", which is inaccurate. Now this warning wouldn't disappear.

Now, the warning would only appear if `foo` points to a record other than `local:foo`, say `provider:bar`, and would mention the canonical id that `foo` points to.

wupr:
> I spotted a few other minor things while looking over this PR which I've also pushed now too.

Thanks! I wanted to change the `get_data` implementation too but wasn't sure whether the new error message was better. But it's definitely good to get rid of the `expect`.

alexrutar:
I see, looks great now, thanks for the fix! Agreed, the error message is now more confusing, but it should in principle never occur unless the database is in an invalid state so I guess it is not such a big deal.

PR #127, opened 2024-11-11
Fix `autobib update` behaviour for local records

wupr:
`autobib update` is not meant to be run on local records, but the current error messages are confusing. For example, if `local:foo` is in the database and `local:bar` isn't, then the current behaviour is the following.
```
$ autobib update local:foo
error: Provider error: Undefined local record: foo
$ autobib update local:bar
error: Record corresponding to 'local:bar' does not exist in database
suggestion: Use `autobib get` to retrieve record
```

Tasks:
- [x] Change the error messages
- [x] Add tests

### DIFF
--- a/src/record/key.rs
+++ b/src/record/key.rs
@@ -190,6 +190,12 @@ impl RemoteId {
         &self.full_id[..self.provider_len]
     }
 
+    /// Check whether the `provider` part of the remote id is `local`.
+    #[inline]
+    pub fn is_local(&self) -> bool {
+        self.provider() == "local"
+    }
+
     /// Get the `sub_id` part of the remote id, after the separator.
     #[inline]
     pub fn sub_id(&self) -> &str {
--- a/README.md
+++ b/README.md
@@ -153,7 +153,7 @@ autobib local my-entry
 ```
 creates a record under the identifier `local:my-entry`.
 You will be prompted to fill in the record, unless you pass the `--no-edit` flag.
-To modify the record later, use the [`autobib edit` command](#modifying-records).
+To modify the record later, run the above command again or use the [`autobib edit` command](#modifying-records).
 
 It is also possible to create the local record from a BibTeX file:
 ```bash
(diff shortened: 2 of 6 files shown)


wupr:
My implementation results in the following now.
```
$ autobib update local:foo
error: Provider error: Unexpected local record 'foo'
$ autobib update local:bar
error: Record corresponding to 'local:bar' does not exist in database
```
Should these two error messages both be changed to explicitly mention that `autobib update` shouldn't be used with a local record? Is it fien to just change the logic in `main.rs` to handle this specific case? Curious about your opinion @alexrutar

alexrutar:
I am starting to wonder if more generally we should be special-casing `RemoteId`? How about a new type called `LocalId` or similar? And adding a third case to `RecordId::resolve`.

I wonder how annoying this would be to implement in the rest of the codebase. In many situations, we definitely do not care, but in some situations, we definitely do. I think `provider error` is quite a confusing error to have, in principle.

Originally, I always imagined that a local id was a relatively rare occurrence; I think I only ever used it once in the previous incarnation of autobib. But as we make support more 'first class' I feel that it could be reflected more precisely in the types.

alexrutar:
After a bit more thought, I think my opinion is to keep the error messages as-is for now, but to keep this in mind for the future.

src/record/key.rs
```diff
@@ -190,6 +190,12 @@ impl RemoteId {
         &self.full_id[..self.provider_len]
     }
 
+    /// Check whether the `provider` part of the remote id is `"local"`.
```
alexrutar, line 193: Do you feel that the quotes are necessary? Maybe `local` is sufficient.

alexrutar (approved):
Looks great to me, only an inconsequential comment on the docstring!

src/record/key.rs
wupr: I guess I wanted to make it explicit this is a string literal, unlike `provider`. But I'll rephrase this.

PR #128, opened 2024-11-14
Add release profile to `Cargo.toml`

alexrutar:
Adds a release profile to `Cargo.toml`. I've added `lto = true`, as suggested for instance [here](https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Practice).

It increases release build compile times a decent amount, but the resulting binary is about 30% smaller. We should definitely use benchmarks to actually decide if it also makes things faster!

@wupr Anything else you think should be added here?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,9 @@ assert_cmd = "2.0"
 assert_fs = "1.1"
 predicates = "3.1"
 
+[profile.release]
+lto = true
+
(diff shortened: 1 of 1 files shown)


wupr:
I'm happy to try LTO and don't have anything to add. I don't think we need to set the opt-level, since it's 3 by default for the release profile.

PR #129, opened 2024-11-22
Add `--rename-from` argument to `autobib local`

### LINKED ISSUE #120: Rename local records
This came up when I was thinking about fixes for people who have `local:::` in their database. Right now the easiest method is probably:
```
autobib get local::: > local.tex
autobib local valid_id --from local.tex --no-edit
autobib delete local:::
```

Fixing this and also typos would be a lot easier if local identifiers can be renamed, e.g.
```
autobib local valid_id --rename :: --no-edit
```

wupr:
Closes #120

### DIFF
--- a/src/db/state/record.rs
+++ b/src/db/state/record.rs
@@ -98,6 +98,29 @@ impl<'conn> State<'conn, RecordRow> {
         Ok(())
     }
 
+    /// Change the canonical id of the row.
+    ///
+    /// Returns `false` if the new canonical id already exists, and `true` otherwise.
+    pub fn change_row_canonical_id(&self, new_id: &RemoteId) -> Result<bool, rusqlite::Error> {
+        let old_id = self.get_canonical()?;
+        debug!(
+            "Changing the canonical id for row '{}' from '{old_id}' to '{new_id}'",
+            self.row_id()
+        );
+        let result = self.prepare(sql::update_canonical_id())?.execute((
+            self.row_id(),
+            Local::now(),
+            new_id.to_string(),
+        ));
+        if let Constraint::Violated = flatten_constraint_violation(result)? {
+            return Ok(false);
+        }
+        self.add_refs_impl(std::iter::once(new_id), CitationKeyInsertMode::FailIfExists)?;
+        self.prepare(sql::delete_citation_key())?
+            .execute((old_id.name(),))?;
+        Ok(true)
+    }
+
     /// Add a new alias for this row.
     ///
     /// The return value is `false` if the alias already exists, and otherwise `true`.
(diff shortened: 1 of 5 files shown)


src/db/state/record.rs
```diff
@@ -98,6 +98,29 @@ impl<'conn> State<'conn, RecordRow> {
...
+        let old_id = self.get_canonical()?;
+        debug!(
+            "Changing the canonical id for row '{}' from '{old_id}' to '{new_id}'",
+            self.row_id()
+        );
+        let result = self.prepare(sql::update_canonical_id())?.execute((
+            self.row_id(),
+            Local::now(),
+            new_id.to_string(),
+        ));
```
wupr, line 114: Does it matter if the execution parameters are owned or borrowed?

src/db/state/record.rs
```diff
@@ -128,13 +151,29 @@ impl<'conn> State<'conn, RecordRow> {
...
+                self.prepare(sql::set_citation_key_fail())?
                     .execute((alias.name(), self.row_id()))?;
                 Ok(None)
             }
         }
     }
 
+    /// Check if the given alias exists and points to this row, and delete the alias if it does.
+    #[inline]
+    pub fn check_and_delete_alias(&self, alias: &Alias) -> Result<(), rusqlite::Error> {
```
alexrutar, line 163: This is a bit of a confusing name, since it doesn't quite reflect that it is relative to the given row? Maybe `delete_if_has_alias` or similar?

src/db/state/record.rs
```diff
@@ -98,6 +98,29 @@ impl<'conn> State<'conn, RecordRow> {
         Ok(())
     }
 
+    /// Change the canonical id of the row.
+    ///
+    /// Returns `false` if the new canonical id already exists, and `true` otherwise.
+    pub fn change_row_canonical_id(&self, new_id: &RemoteId) -> Result<bool, rusqlite::Error> {
```
alexrutar, line 104: Probably this should just be `change_canonical_id`? Since it is a method on `row` already

src/main.rs
```diff
@@ -565,42 +568,71 @@ fn run_cli(cli: Cli) -> Result<()> {
...
+                            row.check_and_delete_alias(&old_alias)?;
+                        }
+
                         let raw_record_data = row.get_data()?.data;
-                        (row, raw_record_data, remote_id)
+                        (row, raw_record_data)
+                    }
+                    RemoteIdState::Null(null_row) => {
+                        null_row.commit()?;
+                        bail!("'{remote_id}' was found in the 'NullRecords' table. A local record should not be present in the 'NullRecords' table.");
```
alexrutar, line 592: Could suggest fixing with `autobib delete --delete-null`

alexrutar (commented):
Looks great, thanks! Just some minor comments

alexrutar:
@wupr After a bit more thought, I realize I am a bit bothered by the name `--move-from` but also I cannot think of good alternatives. I might be slightly happier with renaming `--from` as `--from-bibtex` and then calling this `--from-existing`? But then this is confusing, since `--from-existing` implies that this will copy the existing record...

Maybe `--rename-from`?

src/main.rs
wupr: I thought about this but the `suggest!` would have to come before the `bail!` and that didn't feel quite right. Not sure what's best

wupr:
> Maybe `--rename-from`?

Yep, happy with `--rename-from`!

src/main.rs
alexrutar: Hmm, good point, that's a bit awkward. Because of the way errors are handled a hacky (but equivalent) solution is to just use `error!` and then `suggest!` and then` return Ok(())`

alexrutar: But agreed this feels non-ergonomic.

wupr: Alright, I've implemented the hack-y solution.

PR #130, opened 2024-11-23
Also run tests on other platforms

alexrutar:
Trying out cross compilation to see what happens

### DIFF
--- a/.github/actions/install-mitmproxy/action.yml
+++ b/.github/actions/install-mitmproxy/action.yml
@@ -9,10 +9,27 @@ runs:
       run: pipx install mitmproxy
     - name: Make mitmdump generate certificates
       shell: bash
+      if: runner.os != 'Windows'
       run: mitmdump -s ${{ github.action_path }}/init_cert.py
-    - name: Install CA certificate to system
+    - name: Make mitmdump generate certificates (Windows)
+      shell: pwsh
+      if: runner.os == 'Windows'
+      run: mitmdump -s ${{ github.action_path }}\init_cert.py
+    - name: Install certificate (Linux)
+      if: runner.os == 'Linux'
       shell: bash
       run: |
         openssl x509 -in "$HOME/.mitmproxy/mitmproxy-ca-cert.pem" -inform PEM -out cert.crt
         sudo cp cert.crt /usr/local/share/ca-certificates/
         sudo update-ca-certificates
+    - name: Install certificate (macOS)
+      if: runner.os == 'macOS'
+      shell: bash
+      run: |
+        openssl x509 -in "$HOME/.mitmproxy/mitmproxy-ca-cert.pem" -inform PEM -out cert.crt
+        sudo security add-trusted-cert -d -r trustRoot -k /Library/Keychains/System.keychain cert.crt
+    - name: Install certificate (Windows)
+      if: runner.os == 'Windows'
+      shell: pwsh
+      run: |
+        Import-Certificate -FilePath "$Env:USERPROFILE\.mitmproxy\mitmproxy-ca-cert.pem" -CertStoreLocation Cert:\LocalMachine\Root
--- a/.github/workflows/mitmdump.yml
+++ b/.github/workflows/mitmdump.yml
@@ -16,7 +16,7 @@ jobs:
       cache-hit: ${{ steps.cache-flows-restore.outputs.cache-hit }}
       cache-primary-key: ${{ steps.cache-flows-restore.outputs.cache-primary-key }}
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
       with:
         sparse-checkout: |
           ${{ env.REMOTES_FILE }}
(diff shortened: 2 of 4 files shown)


alexrutar:
Feels like a lot of effort to get all the builds to work, something to think about more later.

alexrutar:
Turns out `aux.rs` is an invalid filename on windows.

alexrutar:
> @alexrutar maybe you have a better idea about the cause.

I have no idea... I don't know how Windows works at all. I was under the impression that the mitmproxy flows files are cross-platform but I could be wrong.

It could be a certificate issue: because the proxy is through localhost the certificate needs to be set up correctly, which is done in the file `.github/actions/install-mitmproxy/action.yml`
```
openssl x509 -in "$HOME/.mitmproxy/mitmproxy-ca-cert.pem" -inform PEM -out cert.crt
sudo cp cert.crt /usr/local/share/ca-certificates/
sudo update-ca-certificates
```
The equivalent code for other platforms is here: https://docs.mitmproxy.org/stable/concepts/certificates/

alexrutar:
We should also setup debug-mode printing to also print the debug formatting of all of the error types!

alexrutar:
My understanding of the sequence is:

1. The `.github/actions/install-mitmproxy/init_cert.py` creates the certificate file at `$HOME/.mitmproxy/mitmproxy-ca-cert.pem` (no idea what the Windows equivalent is)
2. The `.github/actions/install-mitmproxy/action.yml` installs the certificate (currently Ubuntu)
3. The `.github/workflows/mitmdump.yml` generates the flows file (records all of the HTTP requests made while running a command, by proxying through `http://127.0.0.1:8080`)
4. The `.github/workflows/rust.yml` replays directly from the flows file, which requires the certificate to be set up correctly or reqwest will fail to make the connection since the certificate is untrusted.

alexrutar:
Hmm, I found a comment from the Clap maintainer that this type of conflict is not supported:
```
We don't support more specific arg/subcommand conflicts than this. There is a trade off with features we add where they bloat things and make existing ones harder to find. We are considering making the system more flexible but that will mean these extra cases won't be too ergonomic. imo Sometimes its just best to write one off validation after parsing than trying to moded everything in an arg parser.
```

wupr:
> How about let's skip Windows for now and just try to get it to work on MacOS since we both have devices to try that out locally anyway?

Everything works on macOS already. Windows networking was the only problem left. Let me try `certutil` instead of the powershell `Import-Certificate`.

wupr:
> I added debug printing of errors in #194

Ah great!

wupr:
> My understanding of the sequence is:
> 
> 1. The `.github/actions/install-mitmproxy/init_cert.py` creates the certificate file at `$HOME/.mitmproxy/mitmproxy-ca-cert.pem` (no idea what the Windows equivalent is)

PEM just contains string representation of binary data. Should be cross platform.

> 2. The `.github/actions/install-mitmproxy/action.yml` installs the certificate (currently Ubuntu)

I made this cross-platform and all tests worked on macOS.

> 3. The `.github/workflows/mitmdump.yml` generates the flows file (records all of the HTTP requests made while running a command, by proxying through `http://127.0.0.1:8080`)

At first, I just used the flows file generated on Ubuntu. To experiment, I split it to non-Windows and Windows, and the generation failed on Windows. Since you said the flows file should be cross-platform, I've reverted this. 

> 4. The `.github/workflows/rust.yml` replays directly from the flows file, which requires the certificate to be set up correctly or reqwest will fail to make the connection since the certificate is untrusted.

The certificate set-up I wrote yesterday worked on macOS (the command I used was slightly different from the one you quoted from the docs). It didn't work on Windows, and changing the command to the one in the docs didn't fix it either.

wupr:
I'll have to come back to this later. I added a step where tests are run without proxy on Windows. The step gave zbMATH server errors just now, which I remember just happens sometimes? I'll trigger it again later and see.

alexrutar:
> > How about let's skip Windows for now and just try to get it to work on MacOS since we both have devices to try that out locally anyway?
> 
> Everything works on macOS already. Windows networking was the only problem left. Let me try `certutil` instead of the powershell `Import-Certificate`.

Cool, I didn't realize! Good that it's working on macOS. Thanks for all the effort on this, it seems very tedious to get this to work.

> I'll have to come back to this later. I added a step where tests are run without proxy on Windows. The step gave zbMATH server errors just now, which I remember just happens sometimes? I'll trigger it again later and see.

I ran it again and it failed in the same way, which is quite strange, especially since it is currently up for me.

alexrutar:
I have no idea, the request is being correctly proxied.

alexrutar:
ah I totally missed that on windows it is not even running with the proxy! so my guess is that zbMATH is just blocking the request

alexrutar:
My current guess is that things are working fine, since the tests which do not depend on zbMATH succeed with no issue. I'll stop editing this PR now

alexrutar:
I set up the provider-specific success tests with full trace logging; hopefully reqwest will say something more informative about the failed connection.

alexrutar:
> Huh, is mitmproxy not listening on the loopback interface? Or is this a firewall issue?

I have no idea... maybe we can get the `mitmdump` logs to see

wupr:
OK, let me try that

alexrutar:
One more thought, perhaps it makes sense to still build the test binaries on Windows in the PR action? For example this would catch things like accidentally using `aux.rs` as a filename which is not allowed on Windows for whatever reason. Then in the test matrix we can just skip actually running the tests on Windows.

PR #132, opened 2024-11-23
Add `--non-interactive` global flag

wupr:
This simplifies the argument structure of subcommands.

The new global flag
- replaces the `--non-interactive` flag for `autobib edit`;
- replaces the `--no-edit` flag for `autobib local`;
- implies `--prefer-current` for `autobib update`; and
- implies `--force` for `autobib delete`.

Originally proposed in a [comment](https://github.com/autobib/autobib/pull/121#issuecomment-2494960897) on #121.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -439,6 +439,7 @@ fn run_cli(cli: Cli) -> Result<()> {
             let deduplicated = filter_and_deduplicate_by_canonical(
                 citation_keys.into_iter(),
                 &mut record_db,
+                force,
                 |remote_id, null_row| {
                     if !delete_null {
                         null_row.commit()?;
@@ -884,14 +885,15 @@ fn run_cli(cli: Cli) -> Result<()> {
 /// Lookup citation keys from the database, filtering out unknown and invalid remote ids and
 /// undefined aliases.
 ///
-/// Null identifiers are filtered using the provided `null_callback`.
-///
 /// The resulting hash map has keys which are the set of all unique canonical identifiers
 /// corresponding to those citation keys which were present in the database, and values which are
 /// the corresponding referencing citation keys which were initially present in the list.
+///
+/// The resulting hash set contains all of the null identifiers.
 fn filter_and_deduplicate_by_canonical<T, N>(
     citation_keys: T,
     record_db: &mut RecordDatabase,
+    ignore_errors: bool,
     mut null_callback: N,
 ) -> Result<HashMap<RemoteId, HashSet<String>>, rusqlite::Error>
 where
@@ -914,12 +916,20 @@ where
             }
             RecordIdState::UnknownRemoteId(remote_id, missing) => {
                 missing.commit()?;
-                error!("Identifier not in database: '{remote_id}'");
+                if !ignore_errors {
+                    error!("Identifier not in database: '{remote_id}'");
+                }
             }
             RecordIdState::UndefinedAlias(alias) => {
-                error!("Undefined alias: '{alias}'");
+                if !ignore_errors {
+                    error!("Undefined alias: '{alias}'");
+                }
             }
-            RecordIdState::InvalidRemoteId(err) => error!("{err}"),
+            RecordIdState::InvalidRemoteId(err) => {
+                if !ignore_errors {
+                    error!("{err}");
+                }
+            }
         }
     }
     Ok(deduplicated)
--- a/src/main.rs
+++ b/src/main.rs
@@ -13,7 +13,7 @@ pub mod term;
 use std::{
     collections::{
         btree_map::Entry::{Occupied, Vacant},
-        BTreeMap, HashSet,
+        BTreeMap, HashMap, HashSet,
     },
     fs::{create_dir_all, read_to_string, File},
     io::{self, Read},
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -70,6 +70,10 @@ struct Cli {
     #[arg(short, long, value_name = "PATH", env = "AUTOBIB_CONFIG_PATH")]
     config: Option<PathBuf>,
 
+    /// Do not prompt for user action or open the editor.
```
alexrutar, line 73: Is 'open the editor' redundant here? Perhaps it falls under 'user action'. Maybe `Do not require user action`?

src/main.rs
```diff
@@ -136,6 +140,7 @@ enum Command {
         /// The citation keys to delete.
         citation_keys: Vec<RecordId>,
         /// Delete without prompting.
+        /// This is implied by the `--non-interactive` flag.
```
alexrutar, line 143: I don't know if this is a good idea, perhaps `--non-interactive` should just mean "fail without prompting"?

alexrutar (commented):
Thanks for implementing, I agree this is much more ergonomic!

Just a general question, since I haven't tried this out locally: is the placement of the `-y` flag flexible? E.g. can I run both `autobib -y update ...` and `autobib update -y ...`

src/main.rs
alexrutar: Wait this is inside the `delete` command so it's totally fine. Resolving this now.

alexrutar: Actually maybe the help message is a bit long? What does it looks like to run `autobib help delete`?

wupr:
> Just a general question, since I haven't tried this out locally: is the placement of the `-y` flag flexible? E.g. can I run both `autobib -y update ...` and `autobib update -y ...`

We can make it flexible by adding the `global = true` annotation

src/main.rs
wupr: It's a bit long. I'll break it into a new line

alexrutar:
@wupr After sleeping on it, I'm still having second thoughts about the automatic `autobib delete --force`. I might be more inclined to do the following: `--non-interactive` does not imply `--force`, but instead of prompting for confirmation, it write an error to the command line saying that the record could not be deleted because there were multiple referencing keys, and to re-run with `--force`.

alexrutar:
Actually, could we also be a bit more clever?  What if I run `autobib delete key1 key2 key3` where `key1..3` all reference the same record, but there are no other key referencing this record?

I actually realize this also results in a funny bug, that if you say 'yes' to the deletion of `key1` then the command returns an error for `key2` and `key3`! I will fix this in another PR.

alexrutar:
I've just implemented it here for now since I guess it was convenient / I didn't need to wait for the PR to be merged but happy to wait split off into a different PR if you want.

src/main.rs
```diff
@@ -916,12 +918,20 @@ where
...
             RecordIdState::UndefinedAlias(alias) => {
-                error!("Undefined alias: '{alias}'");
+                if !ignore_errors {
+                    error!("Undefined alias: '{alias}'");
+                }
+            }
+            RecordIdState::InvalidRemoteId(err) => {
+                if !ignore_errors {
+                    error!("{err}")
+                }
```
alexrutar: Should we emit this error anyway? For the other two, it is natural that they are suppressed.

alexrutar: This is the error if the Remote Id is invalid

wupr:
> I might be more inclined to do the following: `--non-interactive` does not imply `--force`, but instead of prompting for confirmation, it write an error to the command line saying that the record could not be deleted because there were multiple referencing keys, and to re-run with `--force`.

I think this is very reasonable. But perhaps then we shouldn't use `-y` as the short flag, since `-y` gives the impression that you are saying "yes" to all the prompts. Maybe an alias like `--ni` if we don't have a good alternative?

alexrutar:
Hmm, that's a good point re `-y`. How about `-I` (capital 'i') instead? I wish there was a good reference on flag conventions for command lines.

alexrutar:
From https://clig.dev/#interactivity

> Only use prompts or interactive elements if stdin is an interactive terminal (a TTY). This is a pretty reliable way to tell whether you’re piping data into a command or whether it’s being run in a script, in which case a prompt won’t work and you should throw an error telling the user what flag to pass.
> 
> If --no-input is passed, don’t prompt or do anything interactive. This allows users an explicit way to disable all prompts in commands. If the command requires input, fail and tell the user how to pass the information as a flag.
> 
> If you’re prompting for a password, don’t print it as the user types. This is done by turning off echo in the terminal. Your language should have helpers for this.
> 
> Let the user escape. Make it clear how to get out. (Don’t do what vim does.) If your program hangs on network I/O etc, always make Ctrl-C still work. If it’s a wrapper around program execution where Ctrl-C can’t quit (SSH, tmux, telnet, etc), make it clear how to do that. For example, SSH allows escape sequences with the ~ escape character.

alexrutar:
Also it's a good point that we should be checking for interactivity... but feels like a bigger problem for another day.

src/main.rs
wupr: I think it's fine to suppress it, but I'm happy either way.

wupr:
`-I` is a great idea. Agreed we should be checking for interactivity.

alexrutar:
I've deferred proper handling of `delete-null` for #135 for now

PR #135, opened 2024-11-24
Implement cache eviction

alexrutar:
This PR will eventually implement Cache eviction, as a replacement for `autobib delete --delete-null`.


src/db.rs
```diff
@@ -400,7 +435,7 @@ impl RecordDatabase {
             .query_map([], |row| {
                 if let ValueRef::Text(bytes) = row.get_ref_unwrap(0) {
                     // SAFETY: the underlying data is always valid utf-8
-                    f(unsafe { std::str::from_utf8_unchecked(bytes) });
+                    f(std::str::from_utf8(bytes).unwrap());
```
wupr, line 438: So this adds a check before the conversion. Why is it used in favour of the unsafe block?

src/main.rs
```diff
@@ -312,6 +313,17 @@ enum UtilCommand {
         #[arg(short, long)]
         fix: bool,
     },
+    /// Clear caches based on conditions.
+    ///
+    /// A cache entry is removed if it matches any of the given conditions.
+    Evict {
+        /// Clear entries with citation keys matching this regex.
```
wupr, line 320: Shall we provide a link to the regex specification?

src/main.rs
```diff
@@ -312,6 +313,17 @@ enum UtilCommand {
...
+        #[arg(short, long)]
+        regex: Option<String>,
+        /// Clear cached elements predating the provided time.
+        #[arg(short, long)]
+        before: Option<DateTime<Local>>,
```
wupr, line 325: I think we should make these two options mutually exclusive, since I would usually expect two conditions to be intersected, not union-ed.

src/db.rs
alexrutar: It was for consistency with some other conversions; in principle I think if the underlying database is malformed we should panic rather than cause undefined behaviour. Almost this will not result in any meaningful performance impact (I imagine it takes quite a bit longer to read the data from the SQLite database into memory but this should be benchmarked I suppose)

src/main.rs
alexrutar: Actually I will just implement the 'and' behaviour instead (I thought about doing this but I was a bit lazy)

src/main.rs
alexrutar: Will do!

alexrutar:
I've implemented the `AND` behaviour. One thing which I would have liked is to avoid the `Vec<Box<dyn IntoSql>>` and instead directly use `Vec<&'a dyn IntoSql>` but I could not make the lifetimes work out unfortunately. I guess a `dyn IntoSql` is itself an object (but unsized) and therefore someone needs to have ownership?

src/db.rs
wupr: I see! I was just curious. In general I think I'd prefer `unwrap` as well.

src/main.rs
wupr: Excellent!

alexrutar:
@wupr Feel free to have a go at fixing the lifetimes if you are interested; this would also avoid the extra `Vec<&dyn ToSql>` allocation in `RecordDatabase::evict_cache`

PR #139, opened 2024-12-02
Fix incorrect error code when calling with `-qq`

alexrutar:
To fix this I added a new `fail!` macro, which first sets the error code and then calls the `error!` macro internally.

Not sure if `fail!` is a good name?


wupr:
Ah, well-spotted!

`fail!` might be too similar to `bail!`.
How about just calling it `error!`? We can `pub(crate) use` `warn!` and `info!` in `logger.rs`, so that only that file imports from the `log` crate and we are less likely to accidentally use the uncustomised `error!`. Maybe that's still confusing.

alexrutar:
Hmm I would be amenable to that too, also it means fewer changes to the code

alexrutar:
Well I don't know about fewer changes to the code, but in the future if we decide to add custom hooks to other macros too this makes it much easier.

src/main.rs
```diff
@@ -56,6 +55,7 @@ pub use self::{
     normalize::Normalize,
     record::{get_record_row, Alias, AliasOrRemoteId, RecordId, RemoteId},
 };
+// pub(crate) use logger::{debug, error, info, suggest, trace, warn};
```
wupr, line 58: Comment to be removed?

wupr (approved):

src/main.rs
alexrutar: Good catch, thanks!

PR #155, opened 2024-12-29
Cleanup main

alexrutar:
Implements some cleanup in `main.rs`. Mostly moves things to the new `cli` module, with better organization.


alexrutar:
Going to merge this for now because I want to implement some other stuff but we can always do more cleanup later.

PR #156, opened 2024-12-30
Delay creation of data directory until necessary

wupr:


alexrutar:
What is the error message if a user-provided database file does not have a parent directory which exists?

wupr:
I get this:
```
% autobib --database=~/.typo/records.db util check
error: SQLite error: unable to open database file: ~/.typo/records.db
```

alexrutar:
Is it worth creating the directory for the user automatically in this case? I can see arguments in both directions.

alexrutar:
I think I lean towards making the directory automatically.

I personally find the error message a bit confusing (why couldn't the database file be opened?). Automatically creating the parent directory would mean this error only shows up for more serious reasons. Also `create_dir_all` is quite safe I think.

alexrutar:
Thanks!

Issue #152, opened 2024-12-29
Implement provider auto-discovery


alexrutar:
It could be cool to support automatic discovery of the provider. Maybe this would work in tandem with #144 , i.e. the `preferred_providers` flag; and then if you use a blank provider (e.g. `:1467.28001) and you have something like
```
preferred_providers = ["zbl", "doi"]
```
then we just iterate over `preferred_providers` and the first provider for which the sub-id is valid (only the quick `is_valid` check) becomes the provider for this key.

This is closely tied to the normalization discussed in #44 and probably should be implemented at the same time.

Issue #150, opened 2024-12-29
Implement 'auto aliases'


alexrutar:
We could add a configuration option for 'automatic aliases': one specifies a transformation rule in the config file, like

```
[auto_alias.zbl]
pattern = "..."
```

where `pattern` is a regex with a single capture group, and then the corresponding sub-id is automatically assigned as the sub-id. Some people who are trying out this program like to write `MR....` or `zbMATH...` instead of `mr:` or `zbmath:` and this enables this use-case without friction.

Issue #44, opened 2024-06-14
Implement sub-id normalization


alexrutar:
We could implement key normalization rules so that multiple key formats map to the same underlying key. It is not clear to me if it is better to do this implicitly, or if we should explicitly ask the user to reformat the key (with a suggestion for the canonical format)

Some possible examples of this are:
- arxiv with versions or without versions
- zbmath has or does not have leading `0`
- ISBN should be 13 digits long (auto-convert 10-digit ISBNs)

PR #157, opened 2024-12-30
Implement auto identifier transformations

alexrutar:
The goal of this PR is to provide a general mechanism for 'key normalization', i.e. transformations which are applied to citation keys in order to convert them into `RemoteId`s.

- [x] #44 `provider:sub_id -> provider:new_sub_id`
- [x] #150 `alias -> provider:sub_id`
- [x] ~~#152 `:sub_id -> provider:sub_id`~~

More precisely, this PR implements two changes.

First, the signature for provider validation is changed to return a new
```rust
enum ValidationOutcome {
    Valid,
    Normalize(String),
    Invalid,
}
```
with an implementation of `From<bool> for ValidationOutcome` for convenience. Now, the provider can return a normalized value for the provided key, which will then be used for the underlying remote id. Since we still need to keep track of the original remote id, there is a new `MappedFrom` struct which holds the new remote id, and the original key (if normalization occurred). The `MappedFrom` struct has a custom display implementation which also prints the original key (if any) so the user can keep track of where things are coming from.

Secondly, this implements 'alias transforms'. Implementing this feature involved a quite a few changes.

1. First, since alias transforms are implemented as a list of `(regex, provider)` pairs in the toml configuration, I had to refactor the `Config` implementation to permit an internal `LazyLock`. Unfortunately, providing non-`fn` types for `LazyLock` requires a lot of new type annotations. On the other hand at least now things are somewhat set up if the configuration contains other expensive transformations.
2. The configuration got a new `alias_transform` section, with contains a list of rules for transformations which are applied sequentially, as well as a `create_alias` option to automatically create the actual alias in the database (rather than just performing the transformation).
3. There are two places where alias creation may need to occur: first, the alias could be mapped to a record which already exists in the database, in which case alias creation needs to occur inside `RecordIdState::determine`. Otherwise, the alias could be mapped to a record which does not exist in the database; then, alias creation needs to occur inside `get_record_row_recursive` at the time when remote resolution is completed. Unfortunately I could not see a way to include automatic alias creation in a single location. Because of this, now there is an `ExtendedRecordIdState` enum (needed to pass the relevant context to `get_record_row`), and there is a new `alias_transform` argument added in a lot of places.


src/app/retrieve.rs
```diff
@@ -112,13 +115,13 @@ pub fn retrieve_and_validate_entries<T: Iterator<Item = RecordId>>(
 }
 
 /// Retrieve and validate a single BibTeX entry.
-fn retrieve_and_validate_single_entry(
+fn retrieve_and_validate_single_entry<F: FnOnce() -> Vec<(regex::Regex, String)>>(
```
alexrutar, line 118: As far as I can tell I don't think all of these additional trait bounds can be avoided. The problem is that in order to construct a `LazyLock` which captures from the environment, you need to provide the custom closure type - but closure types are opaque, so we cannot assign a concrete type to the return value of `load_config`. Therefore we need the trait bound everywhere that a `Config` is passed.

alexrutar:
@wupr Since I already made a lot of changes to the configuration in this PR, I also finished handling some of the other issues involving config. In any case, I'm finished editing now; this is a pretty large PR so probably it is unreasonable to read through all of the code carefully but if you take a look most importantly at naming and program behaviour (I think implementation details are less critical) it would be good to sanity check things before committing to any strange behaviour!

src/db/state.rs
```diff
@@ -160,52 +163,170 @@ impl<'conn, I: InDatabase> State<'conn, I> {
...
+/// A representation of the database state beginning with an arbitrary [`RecordId`].
+///
+/// This is an extended version of [`RecordIdState`] which preserves more context required for the
+/// correct implementation of [`get_record_row`](crate::get_record_row).
+#[derive(Debug)]
+pub enum ExtendedRecordIdState<'conn> {
+    /// The `Records` row exists.
+    Existent(String, State<'conn, RecordRow>),
+    /// The `Records` row does not exist and the `NullRecords` row exists.
+    NullRemoteId(MappedFrom, State<'conn, NullRecordRow>),
```
wupr, line 197: Reading this, I think perhaps a better name for `MappedFrom` is `MappedKey`, if the generic type parameter is usually omitted anyway.

wupr (approved):
I have reviewed about 60% of the code, including all function-level changes. Everything seems sensible to me -- I don't want to hold this up so feel free to merge. I saw a bunch of nice optimisation of previous code as well. As always, thanks!

src/app/retrieve.rs
wupr: I agree this is probably unavoidable. (Also if feels like we are dealing with dependent types here!)

alexrutar: Sort of! Though I guess in this case it's more an issue of opaque types, if `fn_traits` were a thing we could specify the concrete implementation as a wrapper type holding the `Vec`. Also I think when https://rust-lang.github.io/rfcs/2515-type_alias_impl_trait.html lands it will also fix it.

src/provider.rs
```diff
@@ -65,10 +65,28 @@ fn lookup_validator(provider: &str) -> Option<Validator> {
     }
 }
 
-/// The outcome of checking that a provider and sub_id are valid.
 pub enum ValidationOutcome {
+    Valid,
+    Normalize(String),
+    Invalid,
+}
```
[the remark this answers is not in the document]
alexrutar: I guess I was thinking (form the 'provider implementer perspective') that there should be a specific type for what the provider returns, which is entirely self-contained? And a provider cannot return "invalid provider"

src/db/state.rs
```diff
@@ -160,52 +163,170 @@ impl<'conn, I: InDatabase> State<'conn, I> {
...
-    NullRemoteId(RemoteId, State<'conn, NullRecordRow>),
+    NullRemoteId(MappedFrom, State<'conn, NullRecordRow>),
     /// The `Records` and `NullRecords` rows do not exist.
-    UnknownRemoteId(RemoteId, State<'conn, Missing>),
+    UnknownRemoteId(MappedFrom, State<'conn, Missing>),
     /// The alias is undefined.
     UndefinedAlias(Alias),
     /// The remote id is invalid.
     InvalidRemoteId(RecordError),
 }
```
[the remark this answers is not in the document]
alexrutar: I kept it because actually in the majority of situations you don't want to distinguish these situations; if anything it's awkward that the extended version is required (exactly once, in `get_record_row`). There are currently 3 other places where the non-extended version is better since the handling of both cases is identical.

But actually it's a good point, it could be a special enum type with a 'flatten' method.

alexrutar:
> I have reviewed about 60% of the code, including all function-level changes. Everything seems sensible to me -- I don't want to hold this up so feel free to merge. I saw a bunch of nice optimisation of previous code as well. As always, thanks!

Thanks again for the review! I had a good time implementing these changes, in the end things feel cleaner on top of there  being new features which is always a good feeling.

src/provider.rs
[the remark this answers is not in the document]
alexrutar: I'm going to leave this one as-is; we can revisit the provider stuff later (e.g. also concerning my comment about making the providers into a separate crate / repository).

I think this is definitely much too early right now but in the future since implementing a provider is somehow entirely independent of all of the other code in principle some more separation could be sensible. But of course the usual issues with circular dependencies (e.g. on our custom `HttpClient`) and also if there are some fundamental architecture changes in the future, e.g. implementing multi-threading or async.

PR #162, opened 2025-01-21
Add `--attachments` mode to `autobib find`

alexrutar:
This method adds a `--with-attachment` flag to `autobib find` to only list items which actually have at least one attachment.

I also moved the 'entry type' printing under a flag (I think it is a bit more natural to not include it by default).

In the future, once `nucleo_picker` merges the 'multiple columns' (https://github.com/autobib/nucleo-picker/issues/1) feature we can actually add the ability to also match against the attachment name, too. But for now I guess this is the next best thing and works for the use-cases that I care about, at least (e.g. "I want the PDF for that paper").

I modified the printed value upon selection with the `--with-attachment` flag to instead print a (list of) attachment paths, since this felt like the most canonical thing to do. However I do not know if this is a good option overall - should there be a specific 'output type' flag? The only thing was that when I added the output type flag, I realized that basically if you  do not call `--with-attachment`, the 'attachment path' output is quite meaningless (if you just want the 'path root' you can forward to `autobib path` yourself); and if you do call the `--with-attachment` it seems quite rare that you actually care about the canonical ID instead of the attachments.

However I am not totally decided about it so I am curious for thoughts here.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,6 +192,7 @@ dependencies = [
  "serde_json",
  "thiserror 2.0.11",
  "toml",
+ "walkdir",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -18,18 +18,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
 
 [[package]]
-name = "ahash"
-version = "0.8.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
-dependencies = [
- "cfg-if",
- "once_cell",
- "version_check",
- "zerocopy",
-]
-
-[[package]]
 name = "aho-corasick"
 version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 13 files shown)


wupr:
The new flags seem very useful. I would name it `--has-attachment` though.

My vote is for a flag for output type. This would make it easier to maintain consistency if we add more filters in the future. Changing the output automatically might surprise the user.

Also it could be useful to select a specific attachment instead of printing out a list?

alexrutar:
Good point I agree that `--has-attachment` is better.

I guess with an output flag, what precisely does it mean to output the attachment path in the case when there are no attachments?

I thought about actually being able to search through attachments but I realized that without the multi-column feature in `nucleo-picker` (i.e. https://github.com/autobib/nucleo-picker/issues/1) it is quite a bit harder to use the search interface in a practical way. Otherwise I totally agree.

Actually maybe this is an alternative proposal: instead we have have 'search modes'; and the mode determines what is selected, and also determines what is printed? And then the `--has-attachment` flag is actually just the 'attachment mode'. Maybe a `--mode {id,attachment,...}`, probably `--mode` is not a good name here but cannot think of something better off the top of my head.

alexrutar:
Hmm one more possibility is to 'double run' the picker: the first run you select the item, and then if there are multiple options it opens up a new picker to choose the precise attachment?

wupr:
Yeah, the two-stage solution is what I wanted to suggest.

Good point regarding the output when there are no attachments.

I didn't quite understand what "multiple columns" meant, but I think I do now. Having search modes would be the better solution, although in this case the flag could just be called `--attachements`.

alexrutar:
Alright I've added both of these features now!

Issue #169, opened 2025-01-28
Add `--skip-from` flag for `autobib source`


alexrutar:
Sometimes, you might have a pre-existing local bibliography file and you only want to generate entries for keys which are not already present in the existing bibliography file.

The `--skip-from` flag would allow skipping records already present in a `.bib` file.

There should also be a related flag `--skip` to which citation keys can be manually passed.

This is quite similar to the append feature described in #103 and should probably be implemented at the same time.

Issue #103, opened 2024-10-23
Add "append"-type argument to `autobib get`


alexrutar:
We should have an "append"-type option for autobib get, which given as input an existing `bibtex` file, finds the keys which are present in the source file (or in the list of `get` arguments) but not in the target bibtex file, and then outputs a "supplementary" bibliography which only contains the new keys.

alexrutar:
Implemented in #170

PR #170, opened 2025-01-28
Implement output filters for `autobib source/get`

alexrutar:
This PR implements #103 and #169. Namely, it adds the following flags to `autobib source`:

- `--skip RECORD_ID`, to ignore a given record id (can be passed multiple times)
- `--skip-from PATH`, to ignore any identifiers present inside path (with the same search logic as used in `autobib source`
- `--append`, used with `--out`, which guarantees that (1) the new entries are appended, and (2) any existing keys already present in the file are automatically skipped (to avoid duplicates)

The `--append` flag is also added to `autobib get` with the same logic; but I did not add the other flags since I feel they are not useful.

One thing which I did not implement is to detect duplicates at the level of 'multiple canonical keys'. I guess this could be relevant for `--append`. My reasons for not doing this were:

1. In my opinion main use-case for `--append` is when there is an existing `.bib` file which contains keys which are coming from some external source. In this situation I would expect that treating existing keys inside the bib file as though they were identifiers in the database is misleading and could result in false positives / weird behaviour.
2. It is annoying to implement since it requires digging quite deep into the retrieval implementation.


wupr:
These all sound sensible to me!

The only things that's different from what I'd have expected is the behaviour of `--skip-from`. I was imagining something like the `-X, --exclude-from=FILE` option of `tar`., where the file is a simple list of patterns to match and exclude. But then I read #169 and I think it makes sense the way you implemented it.

And thanks for understanding I've been quite busy. I'll have a few hours on Saturday to catch up with things, but feel free to make any merges.

src/app/cli.rs
```diff
@@ -247,14 +256,29 @@ pub enum Command {
...
+        #[arg(long, group = "output")]
+        print_keys: bool,
+        /// Skip a citation key (if present).
+        #[arg(short, long, value_name = "CITATION_KEY")]
+        skip: Vec<RecordId>,
+        /// Skip citation keys which are present in the provided `.bib` file(s).
+        #[arg(long, value_name = "PATH")]
+        skip_from: Vec<PathBuf>,
+        /// Override file type detection for skip files.
+        #[arg(long, value_name = "FILETYPE")]
```
wupr, line 280: An underscore between `FILE` and `TYPE`?

src/app/cli.rs
```diff
@@ -247,14 +256,29 @@ pub enum Command {
         /// The files in which to search.
         paths: Vec<PathBuf>,
         /// Override file type detection.
-        #[arg(long)]
+        #[arg(long, value_name = "FILETYPE")]
```
wupr, line 259: Same here

src/app/write.rs
```diff
@@ -9,20 +9,55 @@ use nonempty::NonEmpty;
...
+    append: bool,
     grouped_entries: BTreeMap<RemoteId, NonEmpty<Entry<D>>>,
 ) -> Result<(), serde_bibtex::Error> {
-    if let Some(path) = out {
-        let writer = io::BufWriter::new(std::fs::File::create(path)?);
+    if let Some(file) = out {
+        let mut writer = io::BufWriter::new(file);
+        if append && !grouped_entries.is_empty() {
+            writer.write_all(b"\n")?;
+        }
```
wupr, line 55: Is this in case there wasn't a trailing newline character in the file?

src/app/cli.rs
```diff
@@ -247,14 +256,29 @@ pub enum Command {
...
         file_type: Option<SourceFileType>,
         /// Write output to file.
-        #[arg(short, long, group = "output")]
+        #[arg(short, long, group = "output", value_name = "PATH")]
         out: Option<PathBuf>,
+        /// Append new entries to the output.
+        #[arg(short, long, requires = "out")]
+        append: bool,
```
wupr, line 266: Should we explain that existing keys in the output file will be skipped?

wupr (approved):

src/app/write.rs
alexrutar: Yeah exactly, or even if there is a trailing newline it makes the spacing a bit better in my opinion. The newline works well if you use 'append' multiple times (it will generate a nicely spaced bibliography by default); main downside is if you are using `--append` but the file is already empty.

I also had the idea to check for the number of trailing newlines in the file and only write this if needed but it seemed somewhat unnecessary?

src/app.rs
```diff
@@ -583,54 +600,88 @@ pub fn run_cli(cli: Cli) -> Result<()> {
             paths,
```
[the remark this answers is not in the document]
alexrutar: done!

alexrutar:
@wupr Thanks for taking a look through! I made too many changes in the other PR so I will just merge this one as-is and I've implemented these things on top of the other one.

src/app/write.rs
wupr: I'm happy either way. The spacing makes it easy to see where the appended content starts.

PR #176, opened 2025-04-08
Update dependencies

alexrutar:
Updates some versions and fixes dependabot warning.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -93,15 +93,15 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.97"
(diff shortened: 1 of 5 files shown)


alexrutar:
There is some sort of issue with `jfm:60.0017.02` returning multiple canonical identifiers (it seems like zbmath is trying to be clever and remove duplicates internally, since the same record is also indexed as `zbl:0009.14501`). This will require a bit more work to resolve.

alexrutar:
Anyway I am going to close this now just to get the updated dependencies merged in but I created a new issue and we should handle the situation with `jfm` some time later.

PR #179, opened 2025-08-04
Fix `autobib find` failing to print ID when configuration not set

alexrutar:
The fix is pretty simple, the 'fallback' code was accidentally inside the if branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "adler2"
-version = "2.0.0"
+version = "2.0.1"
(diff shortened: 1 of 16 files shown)


alexrutar:
It seems that arxiv has changed there API some time in the past 6 months...

alexrutar:
For now I've decided to just keep the existing format, and we can re-visit this later if/when we support explicit arxiv versions.

wupr:
I always cite the latest version with the latest revision date. But I also find it frustrating when the the order of publication on paper don't reflect reality...

We could add a field called "first published" to keep the v1 date? And the user could customise the .bib printed (I can't remember whether we already have this functionality).

Issue #140, opened 2024-12-04
Add escapes for invalid characters in error messages


wupr:
Currently,
```sh
autobib -I local "$(echo 'hello\nworld')"
autobib get "$(echo 'hello\nworld')"
```
gives the error message:
```
error: identifier contains invalid character '
'
```

wupr:
This issue also occurs when running `autobib util list` and `autobib info` normally, but perhaps the solution there is to normalise whitespace in keys?

alexrutar:
In the second situation I would actually lean towards not normalizing whitespace? I guess I would consider it to be the user's problem if they decide to do this. I think the invalid character message is a reasonable exception since it makes it more transparent exactly what the issue is.

alexrutar:
Turns out this needed to be fixed upstream; I've done that now (but not posted a new release).

When there is a new `serde_bibtex` version and we update this will be resolved!

Issue #184, opened 2025-08-27
Proper handling and normalization of dates


alexrutar:
There is a bit of weirdness around our handling of dates.

1. Date merging is a bit weird: if `A` has `month = {...}, year = {...}`, and `B` has `year = {...}`, and then we merge `A` into `B` (with priority), we end up with an invalid year since it is mixing month and year fields.
2. We should provide an option to normalize dates, perhaps into the `date = {...}` field? I don't know what to do with incomplete dates.

wupr:
What would happen if the user has previously selected `only-date` and then decide to switch from BibLaTeX to vanilla BibTeX? Should we allow customisation during output as well? Maybe then we only need to keep `none` `date` `ymd` as options?

PR #185, opened 2025-08-27
Fix arxiv implementation, set up database migration, and other small improvements

alexrutar:
Replaces the arxiv implementation with a new crate that I wrote last week ([rsxiv](https://docs.rs/rsxiv)). Lots of benefits:

1. Cleans up the messy XML parsing implementation since `rsxiv` exposes a much cleaner `Deserialize` interface.
2. Now supports handling of arxiv version identifiers correctly.
3. Fixes a normalization bug: the [official arxiv recommendation](https://info.arxiv.org/help/arxiv_identifier_for_services.html) is that old-style identifiers should have the subject class squashed like `math.CA/0003.123` to `math/0003.123`. In fact, internally they support squashing *every* possible thing that matches `[A-Z][A-Z]`. We now perform this normalization as well. Hopefully nobody has identifiers like this in their database, but I will write a migration tool for this.
4. No version now defaults to the most recent version available at the time of request.
5. Sets the date according to the most recent version. See #184 for some more thoughts on dealing with dates. Also we could use the `origdate` field to keep the `v1` date.

@wupr Let me know if you have any thoughts on this!

### DIFF
--- a/src/db/functions.rs
+++ b/src/db/functions.rs
@@ -1,22 +1,13 @@
 use rusqlite::{Connection, functions::FunctionFlags};
 
 /// The available application functions.
-///
-/// Function descriptions:
-///
-/// - [`Regexp`](AppFunction::Regexp): `regexp(re: TEXT, value: TEXT) -> BOOL` returns if `value`
-///   matches the regex defined in `re`
-/// - [`ContainsField`](AppFunction::ContainsField): `contains_field(field: TEXT, data: BLOB) -> BOOL`
-///   returns if `data` (interpreted as record data) contains the provided field
-/// - [`GetField`](AppFunction::GetField): `get_field(field: TEXT, data: BLOB) -> TEXT or NULL`
-///   returns the field value if it exists, or null.
 #[derive(Debug)]
 pub enum AppFunction {
-    /// Checks if a string matches a given regex.
+    /// `regexp(re: TEXT, value: TEXT) -> BOOL` returns if `value` matches the regex defined in `re`
     Regexp,
-    /// Checks if record data contains a field.
+    /// `contains_field(field: TEXT, data: BLOB) -> BOOL` returns if the record data contains the provided field
     ContainsField,
-    /// Returns the field value from the record data.
+    ///`get_field(field: TEXT, data: BLOB) -> TEXT or NULL` returns the field value if it exists, or null.
     GetField,
 }
 


alexrutar:
> Is `[A-Z][A-Z]` effectively the same as `[A-Z]+`? As in they don't have any weird classification strings?

I think this is fine: see https://info.arxiv.org/help/arxiv_identifier.html#identifiers-up-to-march-2007-9107-0703

Also I tried this with the API directly
```
https export.arxiv.org/api/query?id_list=math.Z/0001001
https export.arxiv.org/api/query?id_list=math.ZZ/0001001
https export.arxiv.org/api/query?id_list=math.ZZZ/0001001
```
and the first and third are errors.

alexrutar:
I've implemented the database migration as suggested in #186

However, I have decided against fixing the invalid keys. Retrieving the keys still works correctly, even if the normalization would yield something different, since we always prioritize reading from `CitationKeys`.

The only thing that could happen is accidental duplication of records; but this was already possible with the old normalization (when they were treated as different keys).

alexrutar:
Bumped some versions which now fixes #140

alexrutar:
@wupr To summarize the new changes:

1. Introduced a new `application_id`, which we will use in the future to identify if a given database file belongs to Autobib. This replaces the "check if all the tables look correct" initialization code. Now we no longer check the tables at all. I feel this is fine since (1) it simplifies the code and (2) it improves startup time a bit; but I am curious if you have any objections.
2. The application ID is `0x48a420c4` since this is the first 32 bits of the sha256 hash of the string "Autobib" (try it out with `echo "Autobib" | sha256`). I did some googling and could not find this string or the decimal version of this string anywhere online so it should be safe. (In principle the id should be globally unique.)
3. Refactored the database initialization code to check if the opened database is entirely empty on start (if it does not have the expected application id), so we can safely format it for use with Autobib.
4. Incremented the database `user_version` tag to `v1` and set up a general framework for migration code. The current migration code fixes the constraints (now we have `ON UPDATE CASCADE ON DELETE CASCADE`).

wupr:
Nice! I was also going to propose adding `application_id` after thinking about the database versioning. 

> Do you think this is a good name for the config option? I was also thinking about naming the section [find].

My vote would be for `[find]` over `[picker]`. Just like when I use `fzf` with `fd`, the filtering is done by`fd` as opposed to `fzf`.

alexrutar:
I set up the application-defined function handling in #187; they're currently not used anywhere but I imagine it might be useful in the future. I will stop implementing things here for now since this is getting quite unwieldy.

Issue #153, opened 2024-12-29
Better suggestions for invalid keys


alexrutar:
If a key is invalid (e.g. someone inputs `zbmath:1467.28001` instead of `zbl:1467.28001`) it would be great to provide suggestions.

Maybe the easiest way is for providers to provide these suggestions. Instead of `is_valid` returning a `bool`, it could return a `Result<(), ValidationError` where
```rust
struct ValidationError {
    alternatives: Vec<String>,
    msg: String,
}
```
to catch some common incorrect use-cases.

PR #188, opened 2025-08-29
Add suggested alternatives for invalid keys

alexrutar:
Ads suggestions for invalid keys, as detailed in #153


alexrutar:
I played around with this a bit and the 'database keys' suggestions really led to a lot of false positives, so I think we can just use the 'provider-aware' suggestions only and it will be pretty close. The benefit is that the implementation is now much better (tied directly to the error type).

PR #190, opened 2025-09-01
New version

alexrutar:
This PR cuts a new release `v0.3.0`.

We haven't implemented automated builds / testing on other platforms which was one of my `0.3.0` wish-list items, but at this point I feel a `0.3.0` release is long-awaited so I will advocate cutting one now.

@wupr Can you think of any good reasons to delay?

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -1,9 +1,32 @@
 # Architecture
 ## SQLite database format
+The Autobib command-line tool stores data locally in a [SQLite](https://sqlite.org/) database, located (in order of priority)
+
+- at the location specified by the `--database` command line option
+- as set by the `$AUTOBIB_DATABASE_PATH` environment variable
+- by default at `$XDG_CONFIG_HOME/autobib/records.db`
+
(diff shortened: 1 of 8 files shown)


wupr:
I'd say go ahead. I'm about to be freed from work and could have another go at automated builds this weekend, but I don't think it's worth delaying for that.

alexrutar:
We might want to back-up the original database file when we update the version? Corruption is not a risk since the entire operation is done inside a transaction. However, maybe the user wants to have the old version for roll-back. Especially with automatic updates, maybe this is not desirable in certain circumstances.

I think in most circumstances automatic updates are fine. The only situation I can currently imagine is: I am using `autobib@v0.3.0` and you are using `autobib@v0.2.0`. Under the hood we are using the same shared database file (maybe synchronized online somehow?). I run `autobib@v0.3.0` and it updates the database automatically, and now you cannot read it anymore.

Perhaps I wil upgrate the `info!` on migration to a `warn!`, report that an automatic backup was made, and then continue.

PR #192, opened 2025-09-03
Implement `read-only` compatibility mode

alexrutar:
This PR implements a new `read-only` compatibility mode, which:

- Denies any database writes at the SQL level
- Tries to not make any filesystem writes (not hard-coded but I tried to catch all of the cases)
- Opens the database without trying to migrate or validate it, in "hope for the best" style. A warning is printed that `--read-only` may result in mysterious errors.

To pre-emptively reduce errors, I also added a "command-compatibility" check which as far as I am aware catches all of the cases which could result in a write, and returns an early error.

This also implements custom logic for `autobib get` and `autobib source`, which just fails if the record does not already exist in the database instead of trying to retrieve a new record.


src/app/cli.rs
```diff
@@ -415,6 +415,48 @@ pub enum AliasCommand {
...
+            UtilCommand::Optimize | UtilCommand::Evict { .. } => {
+                Err("Incompatible subcommand with `autobib util`")
+            }
+        }
+    }
+}
+
+impl Command {
+    /// Check if the command is read-only compatible.
+    pub fn is_read_only_compat(&self) -> Result<(), &'static str> {
```
alexrutar, line 433: This is where the compatibility check is manually implemented.

src/app.rs
```diff
@@ -62,6 +65,14 @@ pub fn run_cli(cli: Cli) -> Result<()> {
     );
     info!("SQLite version: {}", rusqlite::version());
 
+    // check for compatibility of the sub-command with read-only mode to try to avoid SQLite
+    // write errors
+    if cli.read_only
```
alexrutar, line 70: This is where the compatibility check is manually implemented.

wupr (commented):
I'll take a look. Off the top of my head, it might be easier to create a `requires_write` group and set it to be mutually exclusive with the `--read-only` flag.

wupr:
Oh I didn't realise we wanted entire subcommands to conflict with the flag. Then given the Reddit comment, I guess what you have is as good as it can be. I see you have emulated the Clap-style error messages.

Some nitpicking on the function naming: `is_read_only_compat` would imply a boolean return value. Maybe `validate_read_only_compatibility`?

alexrutar:
> Some nitpicking on the function naming: `is_read_only_compat` would imply a boolean return value. Maybe `validate_read_only_compatibility`?

Good call, it was returning a boolean before but I changed it at some point! Will fix

alexrutar:
> Oh I didn't realise we wanted entire subcommands to conflict with the flag. Then given the Reddit comment, I guess what you have is as good as it can be. I see you have emulated the Clap-style error messages.

Yeah, I think this is the best possible given the requirements. I am a bit surprised it isn't more common to conflict with full subcommands, but I guess it would be a bit esoteric to setup in a fully generic way.

PR #200, opened 2025-09-14
Remove global option from `--no-interactive`

alexrutar:
Makes `-I/--no-interactive` non-global, to align with `--config` and `--database`. Is there a reason that this was a global flag before?

### DIFF
--- a/docs/changelog/next.md
+++ b/docs/changelog/next.md
@@ -1,0 +1,17 @@
+# Unreleased
+Supported database versions: `<= 1`
+
+Changes since `v0.3.0`.
+
+## Breaking
+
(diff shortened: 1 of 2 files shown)


wupr:
It was a global option for convenience. I would favour making all three global, so one could use them anywhere in the command. Useful when you realise halfway through typing a command that you wanted one of these options.

alexrutar:
Sounds good to me!

alexrutar:
Hmm, setting them global conflicts with some other flags.

What do you think about making them global flags, but setting them to be capital letters in the short form? Then we can stick to the convention that non-global flags are with lowercase letters (except `--help` and `--verbose` which are pretty universal.)

PR #201, opened 2025-09-16
Set up automatic release builds

alexrutar:
Edited to summarize discussion:

Implements a release build script to:

- compile release binaries
- collate dependency licences into a single `license.html` file
- generate a GitHub release
- publish to `crates.io`

You can see a sample GitHub release here: https://github.com/rutar-forks/autobib-test/releases/tag/v0.4.0

This has been tested in https://github.com/rutar-forks/autobib-test and it is working. Release instructions are in [RELEASE.md](https://github.com/autobib/autobib/blob/build-releases/RELEASE.md). There is a helper script in [scripts/release.sh](https://github.com/autobib/autobib/blob/build-releases/scripts/release.sh).

Other changes:
- I added a tag protection rule, which only allows pushing tags to commits for which all of the tests have passed.
- I updated [`README.md`](https://github.com/autobib/autobib/blob/build-releases/README.md) to reflect the new installation instructions. I also modified the 'warning' message a bit to reflect that Autobib is a bit more stable now than it was before.
- I cleaned up the test script and also forced the use of an in-memory database with `--features write_response_cache` to guarantee that the network requests are made.

Some questions:

- [ ] Do we want to support any other targets? Should any of the targets be removed?
- [ ] Any other issues with CI?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,42 +116,30 @@ jobs:
         include:
           - image: ubuntu-latest
             target: x86_64-unknown-linux-gnu
-            use-cross: false
 
           - image: ubuntu-latest
             target: x86_64-unknown-linux-musl
-            use-cross: false
 
-          - image: ubuntu-latest
+          - image: ubuntu-24.04-arm
             target: aarch64-unknown-linux-gnu
-            use-cross: true
 
-          - image: ubuntu-latest
+          - image: ubuntu-24.04-arm
             target: aarch64-unknown-linux-musl
-            use-cross: true
 
           - image: macos-latest
             target: aarch64-apple-darwin
-            use-cross: false
 
           - image: macos-latest
             target: x86_64-apple-darwin
-            use-cross: false
 
           - image: windows-latest
             target: x86_64-pc-windows-gnu
-            use-cross: false
 
           - image: windows-latest
             target: x86_64-pc-windows-msvc
-            use-cross: false
 
-          - image: windows-latest
+          - image: windows-11-arm
             target: aarch64-pc-windows-msvc
-            use-cross: false
-
-    env:
-      CARGO: cargo
 
     steps:
       - uses: actions/checkout@v5
@@ -161,17 +149,6 @@ jobs:
         with:
           toolchain: ${{ env.RUST_VERSION }}
           targets: ${{ matrix.target }}
-
-      - name: Install cross
-        if: matrix.use-cross
-        uses: taiki-e/install-action@v2
-        with:
-          tool: cross
-
-      - name: Overwrite build command env variable
-        if: matrix.use-cross
-        shell: bash
-        run: echo "CARGO=cross" >> $GITHUB_ENV
 
       - name: Install build deps (linux-musl)
         if: matrix.target == 'x86_64-unknown-linux-musl' || matrix.target == 'aarch64-unknown-linux-musl'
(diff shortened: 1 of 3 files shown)


alexrutar:
I ran a ton of trials in https://github.com/rutar-forks/autobib-test and it is working now.

Basically, the steps for building are as follows:

- Update `Cargo.toml` version.
- Rename `docs/changelog/next.md` to the version tag, like `docs/changelog/v0.4.0.md`.
- Wrap this in a PR, wait for check to pass, and merge.
- Pull the commit,  and create an annotated tag `v0.4.0` (the annotation does not matter).
- CI will automatically run and create a draft release containing everything requred.
- Browse into the 'releases' and create the release.

alexrutar:
Maybe we can use https://github.com/onur/cargo-license like
```
cargo license --authors --do-not-bundle
```

alexrutar:
I've set things up with cargo about: https://github.com/EmbarkStudios/cargo-about

wupr:
Will review by Friday!

alexrutar:
> Will review by Friday!

Sounds great! No rush!

alexrutar:
I've pinned the Rust version now to avoid the (unlikely, but possible) scenario where CI runs, and then a release is cut in between a version change.

.github/workflows/release.yml
```diff
@@ -0,0 +1,252 @@
...
+            target: x86_64-unknown-linux-gnu
+            use-cross: false
+
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-musl
+            use-cross: false
+
+          - os: ubuntu-latest
+            target: aarch64-unknown-linux-gnu
+            use-cross: true
```
wupr, line 126: Do we need `cross`? Did we try Cargo-native cross-compilation from x86 to ARM?

wupr (approved):
Reviewed from my phone but all looks great! Thanks!

Shall we also do ARM64 MUSL and x86 macOS?

A stylistic suggestion that just occurred to me. `echo >&2 [message]` might be easier to read than `echo [message] >&2`. Similarly with `xargs < [file] -- ...`.

wupr:
Oh, is the MSVC target statically linked? Maybe we should remove the MSVC targets and add an ARM64 GNU target on Windows?

alexrutar:
Looks like cross is unavoidable (without a lot of fiddling)

alexrutar:
I set up trusted publishing (https://crates.io/docs/trusted-publishing) and also created a `release` environment which means that the workflow can only be run in `main` and also requires manual approval by either of us. This was the recommended way to set things up.

wupr:
Not sure what I was thinking when I said removing the MSVC targets, but maybe we should still add a compilation flag or Cargo.toml config that does static linking? I couldn't find a source that says Windows 10/11 always has the runtime. There is a Universal C Runtime shipped by default, which is distinct from the MSVC runtime?

Happy with everything else! @alexrutar

alexrutar:
> Not sure what I was thinking when I said removing the MSVC targets, but maybe we should still add a compilation flag or Cargo.toml config that does static linking? I couldn't find a source that says Windows 10/11 always has the runtime. There is a Universal C Runtime shipped by default, which is distinct from the MSVC runtime?
> 
> Happy with everything else! @alexrutar

Sounds good! Static linking is reasonable to me, I understand that Rust uses as little as possible of the MSVC C runtime anyway so it should impact binary size in a minimal way. I spent way too much time googling and the main conclusions I found are that (1) The MSVC runtime is basically always available, but (2) Statically linking is probably the best idea anyway

alexrutar:
CI is failing since mathscinet is down, I'll come back to this later

alexrutar:
Hopefully CI starts to succeed later - anyway now Cross has been removed since Github provides native ARM runners for Windows / linux! I also added the tests for the other runner OS platforms.

PR #214, opened 2025-10-01
Autobib is now available via Homebrew tap

wupr:
This PR adds instructions for installation via Homebrew to README.md and makes other release-related changes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -80,8 +80,8 @@ jobs:
       - name: Install cargo-about
         shell: bash
         env:
-          CARGO_ABOUT_VERSION: 0.8.0
-          SHA256: "b53515a1c85f37810ae03ed03a984cde60282eb1fba08ad0888bd9a62f7d5629"
+          CARGO_ABOUT_VERSION: 0.8.2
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yml
```diff
@@ -80,7 +80,7 @@ jobs:
       - name: Install cargo-about
         shell: bash
         env:
-          CARGO_ABOUT_VERSION: 0.8.0
+          CARGO_ABOUT_VERSION: 0.8.2
           SHA256: "b53515a1c85f37810ae03ed03a984cde60282eb1fba08ad0888bd9a62f7d5629"
```
alexrutar, line 84: This is probably going to not work because the SHA256 should be updated. I think the correct value is
```
797bee53a904aab73b31c923da521b72ed5d4ac950903c928f6be3ed236de8b2
```
but should double check from here: https://github.com/EmbarkStudios/cargo-about/releases/tag/0.8.2

alexrutar (approved):
Looks great to me! Thanks for cleaning up the workflow file as well. It would be best to try to run the release (in a fork) to confirm that it works.

alexrutar:
If you rebase on top of `main`, CI should succeed.

PR #223, opened 2025-10-04
Release version 0.4.1

alexrutar:
We need a new release because zbMath is broken.

I'm not 100% sure that the new implementation will always work, but better 90% functional than 0% functional.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -115,7 +115,7 @@ dependencies = [
 
 [[package]]
 name = "autobib"
-version = "0.4.0"
+version = "0.4.1"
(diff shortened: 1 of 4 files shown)


alexrutar:
Thanks for handling homebrew! The release is live now.

alexrutar:
Let me know that you're fine with this (or if you have a better solution).

I will add a step to the release workflow to check that the current branch is 'main'.

alexrutar:
I guess it's also fine to just leave it, `v0.4.1` just points somewhere funny but everything will still work.

PR #226, opened 2025-10-05
Run Markdown lints in CI

wupr:
See https://github.com/DavidAnson/markdownlint for rule definitions and configuration schema. See https://github.com/DavidAnson/markdownlint-cli2 for the command line tool that corresponds to the GitHub Action used.

When fixing existing lint errors, I disabled two rules in `.markdownlint.yaml` and made an ad-hoc exception in `README.md:291`. I didn't go through the entire list of lints and check if there were anything else to disable.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,10 +1,10 @@
 name: CI
-description: Run Rust tests and checks
+# Run Rust tests and checks
 
 on:
   pull_request:
   push:
-    branches: [ "main" ]
+    branches: ["main"]
 
 permissions:
   contents: read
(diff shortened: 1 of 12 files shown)


alexrutar:
Is there an easy way to run this locally / is there a tool to perform these fixes for me?

I wonder if we could make a docker/podman image which includes all of the dev testing tools (like shellcheck, etc.) and use that inside `scripts/test.sh` instead.

wupr:
> I wonder if we could make a docker/podman image which includes all of the dev testing tools (like shellcheck, etc.) and use that inside `scripts/test.sh` instead.

That's a good idea; I'll open an issue on this.

`markdownlint-cli2` is the first JavaScript tool we are using (and I didn't find other good alternatives). I prefer Deno to Node.js, so I run
```sh
deno run --allow-read --allow-sys --no-config npm:markdownlint-cli2 '**/*.md'
```

I'll merge this PR if you are happy with it.

alexrutar:
Looks good to me! Happy to merge as is.

PR #228, opened 2025-10-05
Implement `--template` support and support it with `autobib find`

alexrutar:
This PR implements basic template support for `autobib find`. It also lays out the basic support components for template support in other commands (e.g. with `--rename`).

Try it out with something like
```sh
autobib find -t '{author}: {title}'
```

I tried to make a minimal fully-functional implementation. In particular, this does not support any 'function-like' features (e.g. converting an author list to a list of last names, etc.). See the [syntax docs](https://github.com/autobib/autobib/blob/implement-find-format/docs/template.md) for what is supported.

I also added a `find.default_template` config value, for specifying a default template to be used. The default value here is
```txt
{author} ~ {title}{=subtitle ". "}{=subtitle subtitle}
```

The main user visible breaking changes are:

1. The `--fields` flag has been removed.
2. The `--all-fields` flag has been replaced with a `--strict` flag; the behaviour is somewhat different. Right now, `--strict` mode is only being used to perform field key expansion filtering, but I did not give it a more precise name since in the future we might want it to also mean other things, like failing to render if a function call is not successful, etc.

Template processing is implemented using [`mufmt`](docs.rs/mufmt). I tried to make it as performant as possible:

1. The template is parsed into an intermediate representation immediately, which (1) helps to catch syntax errors, and (2) provides a more efficient rendering representation and avoids repeated compilation.
2. Since template rendering could contain keys which are not in order, we cannot use the previous 'sorted iteration' approach. Instead, I wrote a few heuristics to try to guess the best way to render; these heuristics are computed at template render time. For small templates (<= 4 field keys) I think it is fastest to just brute-force search, i.e. iterate 4 times in the worst case. For templates in which the keys happen to be sorted, we use the previous synchronized iteration approach. Finally, for examples with many keys, we allocate a temporary borrowed `RecordData` type, which holds the `RawRecordData` key/value pairs. I did not run any benchmarks so I have no idea if 4 is a good number here.

A performance optimization which I did not include is one which requires some API changes to the `nucleo-picker/nucleo` crates which would avoid the intermediate `String` representation when rendering. This needs more upstream work / thought to determine the best solution.


alexrutar:
I implemented pretty-printing for error messages with clap. It's a bit hard-coded but looks great... try e.g.
```
cargo run -- find -t 'invalid {%2} '
```

alexrutar:
By the way, I have an extremely strange bug (seems like xcode related somehow?). To reproduce, try running with
```
cargo run -- find -t '{invalid'
```
If you get a crazy error message containing 'CFPropertyListCreateFromXMLData' let me know and I will report it upstream in cargo...

docs/template.md
```diff
@@ -0,0 +1,102 @@
...
+Autobib commands which accept templates also accept a `-s/--strict` flag.
+When used, this flag only formats if all of the field keys which would be rendered actually exist in the Bibtex data.
+
+The precise behaviour depends on the command:
+- `autobib find`: Any record missing a key is omitted from the search interface.
+
+Strict mode applies in two cases:
+
+1. Given a basic expression `{key}`, if `key` is not present.
+2. Given a conditional expression `{=key1 key2}`, if key1` is present and `key2` is not present.
```
wupr, line 89:
```suggestion
2. Given a conditional expression `{=key1 key2}`, if `key1` is present and `key2` is not present.
```

docs/template.md
```diff
@@ -0,0 +1,102 @@
...
+
+The default behaviour can be manually enforced for specific expressions in strict mode by writing `{=key key}` in place of `{key}`.
```
wupr, line 91: It bothers me slightly that the behaviour depends on if `key1` and `key2` are the same. How about `{key?}` for escaping strict mode?

alexrutar:
I've rebased, and also I updated `scripts/test.sh` to require deno to run the markdown lints. Deno seems very nice, usually I avoid running anything JS-related locally because I refuse to install node on my computer for security reasons.

alexrutar:
As far as I can tell, I am done with this PR now.

Issue #246, opened 2025-11-09
piping behaviour


aris-mav:
Something like `autobib get (autobib find)`  (looks for an entry, gets the bibtex information) works very well in my fish shell.

I would expect `autobib find | autobib get` to result in exactly the same behavior, but instead it gives:

```
 autobib find | autobib get
An unexpected error occured while running the program:
  1. Your database file could be malformed or has been edited by another program.
     Run `autobib util check` to see if this is the case.
  2. If you have ruled out 1., this is a bug in autobib. Please report it at
     > https://github.com/autobib/autobib/issues
     including the error message below and any other information you can provide
     about the context in which it occured.

The following is a description of the error which occured:

panicked at library/std/src/io/stdio.rs:1165:9:
failed printing to stdout: Broken pipe (os error 32)
```

I imagine most people would be used to working with pipes in similar tools, so it'd be nice if such behavior could be implemented.

aris-mav:
Sounds good! Let me know in case I can help in any way.

aris-mav:
Just built from that branch and I confirm it works well. Thanks!

Issue #15, opened 2023-12-20
Improve networking / Async


alexrutar:
This will be a long running issue, since I think it's quite hard and touches on a lot of implementation details. We need to handle:

- [x] User identification, especially with things like Crossref that want the API requests to include user / agent data in the header.
- [ ] Rate limitations, and properly handling server requests to stand off.
- [ ] Async.
- [x] ~~Request consolidation, e.g. for arxiv.~~

Fixing these things will require substantial modifications to the `source` implementation. I think Async and Request consolidation could play well together, since the source can just freely make requests whenever it wants something, and wait for the response. But then it needs to be set-up to assume that a single response could contain the data for a large number of individual requests, so rather than there being a pairing (`source:sub_id -> response`) it has to be a pairing (`Vec<source:sub_id> -> response`) and then then Source needs to communicate the grouping mechanism to the network implementation, etc...

Anyway this issue is just a general location to included all of the details that should be worked out at some point.

As an intermediate step, I think we should just create a single `reqwest::blocking` client and pass it to each of the Source. Each Source can independently make any HTTP request that it wants, and then things like user identification, etc. can automatically be included in every request.

alexrutar:
Couple comments:

1. The [sqlx](https://docs.rs/sqlx/latest/sqlx/) crate is probably a good alternative to rusqlite for async support.
2. This was discussed somewhere else, but anyway maybe a `dyn Trait` approach will be more flexible in the future than the current current "static dispatch on function pointer" implementation.

alexrutar:
Anther option in the async database space: [toasty](https://tokio.rs/blog/2024-10-23-announcing-toasty)

wupr:
A comment on the progress indication. Some terminals like GNOME Terminal and Ghostty support control sequences for displaying a progress bar/pie in the GUI.

PR #247, opened 2025-11-09
Support standard input with `autobib source`

alexrutar:
You can now pass files directly through stdin with `autobib source`: for example, `cat file.txt | autobib source`. The implementation is naive and just loads all of stdin into a buffer. Quite bad if stdin is huge. This should be improved in the future, when streaming input is supported (i.e. #15 , and friends).

The file type defaults to `.txt`, which means if you are reading from input which does not just consist of one identifier per line, then you need to explicitly pass `--file-type` like `cat file.tex | autobib source --file-type tex`. But if you have the file, one should just use `autobib source file.tex` directly.

It fixes part of #246; the broken pipe error message should be handled separately which will be more work.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -10,7 +10,7 @@ mod write;
 use std::{
     collections::{BTreeSet, HashSet},
     fs::{File, OpenOptions, create_dir_all, exists},
-    io::{Read, Seek, copy},
+    io::{IsTerminal, Read, Seek, copy},
     iter::once,
     path::{Path, PathBuf},
     str::FromStr,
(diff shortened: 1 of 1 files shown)


wupr:
To further match the behaviour of `cat`, we could have `autobib source` read from stdin automatically if no paths or hyphens are supplied at all (and ignore stdin if there are paths but no hyphens)

alexrutar:
But defaulting to `.tex` input causes issues too: in this case, `cat remotes.txt | autobib source` would just silently succeed and print nothing!

Maybe the best solution is to require `--file-type` if `-` is passed.

PR #253, opened 2025-11-11
Handle broken pipe errors

wupr:
As reported in #246, Autobib may try to write to a broken pipe and crash. This is because Rust ignores SIGPIPE by default instead of terminating the program with exit code 141 (the latter is the expected behaviour of a typical command line program on a Unix-like OS). See discussion at rust-lang/rust#62569 and [two solutions on StackOverflow](https://stackoverflow.com/a/65760807).

This PR implements the solution that singles out broken pipe errors and makes the program terminate early. It replaces the infallible `println!` with a fallible custom macro `owriteln!` and bubbles IO errors to the main function -- except broken pipe errors, which a wrapper of the stdout lock handles immediately by terminating the program.

### DIFF
--- a/docs/changelog/next.md
+++ b/docs/changelog/next.md
@@ -1,1 +1,14 @@
 # Unreleased
+
+Supported database versions: `<= 1`
+
+Changes since `v0.5.0`.
+
+## Other changes
+
+- When prompting for user input, the prompt text is now printed to stderr instead of stdout.
+
+## Fixes
+
+- Fixed a bug where the program panicked on broken pipe errors when writing to stdout.
+  Now the program terminates silently on such errors.


src/term/input.rs
```diff
@@ -14,9 +16,12 @@ impl<S: Display> Input<S> {
     }
 
     pub fn input(&self) -> Result<String> {
-        let mut stdout = stdout();
-        write!(stdout, "{}: ", self.message)?;
-        stdout.flush()?;
+        {
```
alexrutar: Does this code ever run in a non-interactive setting?

alexrutar: Hmm, should this instead be happening in `stderr`? Most interactive text does; for example, the output of `autobib find`.

alexrutar:
Fair enough! I am happy for it to be merged in the current state.

src/term/input.rs
wupr: Sounds sensible. I've switched it to `stderr`

PR #254, opened 2025-11-11
Upgrade rsxiv to fix arxiv API issues

alexrutar:
arXiv decided that they would change their API so I had to fix the response parsing.


alexrutar:
CI broke since arXiv updated their API between when the PR cache was created, and when the `main` cache was created.

PR #256, opened 2025-11-12
Release version 0.5.0

alexrutar:
Because of the arXiv API changes, I am advocating for a new release. Plus, this will break up the changes a bit, since the database refactoring will ultimately introduce a number of API changes which we can put in the next minor version.


wupr:
Good call!

Issue #252, opened 2025-11-10
Determine provenance of `local:` identifiers


alexrutar:
It would be nice to be able to "add provenance" to local identifiers. Something like a argument `autobib provenance local:xyz`, which is similar to the behaviour of `autobib update`, but unlike `autobib update` (which requires a remote provider) only accepts `local:...` and tries to determine a remote provider automatically using the same logic as #251 

I am undecided if this should be a separate function, or if it actually belongs inside `autobib update`. Maybe `autobib update local:xyz` should just work like this automatically under the hood; the behaviour is non-overlapping anyway.

aris-mav:
> Maybe autobib update local:xyz should just work like this automatically under the hood; 

Second that, sounds more intuitive like this, and less commands to memorize (unless of course the `provenance` method would do other things as well).

alexrutar:
This has now been implemented in #257 (the PR is basically complete, though I want to get some local usage out of it for a while before merging it).

I have implemented this under `autobib dedup`, since after some thought it seems to me that this is a special type of deduplication which checks for equivalence of records using identifiers found within the record data.

Issue #251, opened 2025-11-09
Change `autobib import` identifier defaults


alexrutar:
As briefly discussed in #245 it is probably clearer if `autobib import` tries to determine canonical identifiers from the bibtex data by default.

Perhaps the best solution is for `autobib import` to by default attempt to be as clever as possible and determine a nice canonical identifier for each entry. For `doi`, `zbmath`, and `zbl`, it seems that these can be keys found like:
```
 doi = {10.4171/JEMS/1283},
 zbMATH = {7757003},
 Zbl = {1561.28032}
```
For arXiv, they use:
```
archivePrefix = {arXiv},
eprint = {0902.0885},
```
For mathscinet, it is
```
MRNUMBER = {4968626},
```
Not sure about others.

alexrutar:
> > Not sure about others
> 
> ISBN perhaps?

Good catch! Absolutely

alexrutar:
If you import an entry like
```
@book{my-entry,
   title = {...},
   doi = {...},
}
```
And it converted it to `doi:...` would you expect the alias to be created automatically as well (i.e. `my-entry` being an alias which points at `doi:...`)? Also, did you find that the alias documentation is unclear?

aris-mav:
> > On a very similar note, if I'm not mistaken, LibreOffice uses some sort of CSV format to import bibliography information to be used as citations. Maybe the `autobib export` could also offer a `--csv` flag to accommodate that? Could be part of [#158](https://github.com/autobib/autobib/issues/158) ?
> 
> See [#158 (comment)](https://github.com/autobib/autobib/issues/158#issuecomment-3511736675) [fixed incorrect link]
> 
> Can you link some documentation for what LibreOffice expects? In [#158](https://github.com/autobib/autobib/issues/158) , not here.

Quick update on this, I was going though the documentation for libreoffice recently and did not manage to find anything. Might have been AI slop that I came across previously.

aris-mav:
> A small neovim lua plugin utilizing telescope to fuzzyfind outputs of all the autobib commands sounds straightforward enough.


> This would probably require some changes on the Autobib side as well, to generate the match list but not inside the Autobib-native picker interface.

Hey, just a random update on this, thought might be useful to share.
While working on something else, I came up with the following solution for tmux:

In the tmux config, you can `bind -n M-a run-shell "~/.config/tmux/autofind.sh"`, where `autofind.sh` is
```
#!/bin/sh

[...]

Issue #249, opened 2025-11-09
Also import files


alexrutar:
As motivated by #245 : `autobib import` should support file imports. If the biblatex file contains
```
  file = {/path/to/file}
```
then this data should be removed from the entry data and the corresponding file should be copied to the canonical file location supported by `autobib path`.

This should probably be behind a flag; like `--include-files` or something more sensible.

Issue #233, opened 2025-10-07
Support changelog features


alexrutar:
We've been diligently maintaining a changelog; it would be great to be able to use it!

## `autobib hist`
The simplest command is perhaps `autobib hist <record_id>`. This should print a list (in descending order chronologically), starting with the record itself, followed by all previous versions of it. In order to implement this robustly I suspect we will need to update the Changelog a tiny bit. I think merges are not correctly tracked in the current version.

## `autobib undo/redo`
The next command is to support `autobib undo`. This would read the last entry in the Changelog, and perform an action which undoes it. However now there is more complexity: do we 'pop' the entry from the Changelog? (If so, then we cannot 'undo' an 'undo')? Or do we insert a special 'Undo' entry? If we insert a special 'Undo' entry, what happens if you call Undo again?

Logically I think 'undo' and 'redo' are correctly implementation in terms of tree-navigation (Undo moves towards the root, Redo moves towards the leaf, unless there are multiple options in which case the user needs to choose which route to follow). So the Changelog list is quite a bad data structure for this.

alexrutar:
Here's a sketch of possible implementation. It has a lot of rough edges but maybe there are some good ideas inside it. Notably I don't know how to handle `Merge` correctly.

## Internal logic updates
1. The Records table is append-only and contain a new column called 'Parent', which is either null or contains the row-id of another row in the Records table. We also add a special 'Deleted' entry, which has data which is null / empty.
2. If a record is updated, we just add a new row with the new data, adjust the `CitationKeys` ref to point to the new data, and then insert the 'Parent' ref to the previous row.
3. If a record is deleted, we do the same but with the special `Deleted` entry into `Records`. I.e. a 'delete' is like 'update with null'.
4. If records are merged, this is registered as an Update on the target, and a Delete on the other data.

## Tracking merges
In order to keep track of merges and undo them, we would create a new 'Merged' table which contains all of the canonical ids which were merged into another canonical id (i.e. one column with the canonical-id, another column pointed to the deleted row, and a third column pointing to the row it got merged into).

Note that the 'Parent' could be a 'deleted' row - for example, if you delete a record, and then insert a new record with the same name.

## Examples
1. Now the `undo` operation is: update the pointer in `CitationKeys` to point to the parent.
2. The `redo` operation: look for all rows which have the current entry as a parent. If there are multiple, say that the future is diverged and prompt to pick from one of the branches.
[...]

PR #257, opened 2025-11-12
Internal database reimplementation

alexrutar:
This PR re-implements a lot of the database internals. Most notably, canonical identifiers are no longer unique elements of the database: instead, each canonical identifier corresponds to a _tree_ of rows in the database, and each row contains the row-ids of the parents and children (if any).

The entries in the `CitationKeys` index still point to unique row-ids. There is a new invariant: only one row can be active at a given moment (i.e., all elements in `CitationKeys` pointing to a given canonical id must also point to the same row).

The reason for this change is to implement support for undo/redo; i.e. #233 Now, undo and redo are first-class and simply amount to changing the row ids in the `CitationKeys` table to point to a different entry in the same tree. Undo always works (unless you are at the root) and redo will require a prompt if there are divergent changes.

So far, the following additional changes/simplifications have been made.

1. The implementation of `autobib delete` is a lot simpler. We just update a given node to replace it with a 'deletion' marker. This is "soft deletion". We no longer delete the elements in `CitationKeys`. If an attempt is made to retrieve a soft-deleted entry, the appropriate methods will report an error. There is also `autobib delete --hard`, which evicts all mention of a given key, deletes aliases, etc. Also, we no longer worry about deduplicating the inputs (this is what is expected, e.g. with `rm`: the first deletion will succeed, and then subsequent ones will return errors).
2. `autobib merge` has been removed. Instead, there is `autobib delete --replace <key>`, which soft-deletes a record and indicates that it should be replaced by another record. Attempting to retrieve a soft-deleted key in this way will return an error message which suggests a replacement key.
3. `autobib update` has been changed so that the update can be performed in one of three ways: from the remote data, from a bibtex file using `--from-bibtex`, or from another reford using `--from-key`. The `--from-bibtex` argument was previously called `--from`. I also fixed it so that data provided by a from a bibtex file or from another record cannot be used to update an entry in the `NullRecords` table.
4. `autobib undo` and `autobib redo` have been implemented. Since every state has a unique parent, `autobib undo` requires no new information to succeed. In contrast, `autobib redo` requires extra information since each state can have multiple children. Therefore, `autobib redo` accepts an optional `INDEX` argument, which provides the index of the divergent change. If `INDEX` is not provided and there is exactly one child, we can redo to that child anyway; otherwise, `INDEX` is required.
5. A new command `autobib log` has been added. By default, it only shows the immediate history of the active version, but it also accepts a `--all` argument, which shows the full history of a given identifier, including divergent changes.
6. I added a missing `NOT NULL` constraint in the `CitationKeys` table.

I will update this list as this PR continues.

TODOs:

- [x] Implement database migration
  - [x] The 'Records' table migration is quite simple, since the new table has default values for the new columns.
  - [x] Unwind previously merged providers (i.e. canonical ids in `CitationKeys` which point to a different canonical id in `Records`), and replace them with entries equivalent to `autobib delete --replace`.
  - [x] Rename the `CitationKeys` table to the new name `Identifiers`; there is a new NOT NULL constraint so it requires a bulk re-insert.
  - [x] Delete the `Changelog` table, which is no longer used at all.
- [x] Update internal documentation, especially adding docs about the new state representation
  - [x] Check for naming consistency in help messages
  - [x] Add `info!` and `debug!` logging where appropriate
- [x] Implement #252; this should be a two step (1) compute provenance and (2) essentially run `autobib delete --replace`
- [x] Fix the implementation of `autobib util check`
- [x] Update changelog
- [x] Add tests for the new behaviour
- [x] Change `autobib delete`
- [x] Remove `autobib merge`
- [x] Change `autobib update`
  - [x] Apply normalizations on incoming remote data for `autobib update`
- [x] Implement `autobib undo` and `autobib redo`
- [x] Implement `autobib reset`
  - [x] Reset to a specific revision
  - [x] Reset to a specific timestamp
- [x] Better logging for `undo` and `redo` and `reset`, printing information about the new state after the undo
- [x] Correctly handle `autobib update` and `autobib edit` when targeted towards a 'deletion' record
  - [x] This adds a `autobib update --revive` option to update even if the record does not exist. I've also added a `autobib hist revive` command as a general command for adding new versions of existing records. The difference between `autobib update --revive` is that it retrieves data from remote, whereas `autobib hist revive` asks you to input the data manually. I think `autobib edit` should just fail if the data does not exist.
- [x] Reconsider `autobib:local` automatic identifiers
  - [x] While cleaning up the code, I have temporarily removed the automatic alias creation associated with local identifiers. Is this reasonable?
- [x] Changes to `autobib edit`
  - [x] If a normalization is specified, no longer open the editor.
  - [x] Add headless edit methods.
- [x] Correctly handle files, if files exist on soft deletion. Perhaps `autobib find --attachments` should also include soft-deleted entries? Maybe warn if files exist, and say the user should deal with them? Differences between soft / hard deletion? Since merging no longer changes the canonical identifier, we no longer need to move files around. (probably defer this to a new PR)
- [x] Implement #249 (probably defer this to a new PR)
- [x] Implement #251 (probably defer this to a new PR)
- [x] Add ways to prune old entries (defer?)
  - [x] Delete non-active versions which are older than a given age
  - [x] Only keep `n` most recent changes (plus their children)
  - [x] Prune void entries (deleting if present)
- [x] Add new alias subcommand `autobib alias update`
  - [x] Added `--update-aliases` option which deletes/renames the aliases depending on whether `--replace` is specified
- [x] Add 'global' changelog operations
  - [x] Showing all recent changes, up to some limit (count, after time, ...)
  - [x] Bulk options for `autobib reset`, for example to undo bad imports.
- [x] Add a new command `autobib log`
  - [x] Provide option show the history / sequence of parents of the current node.
  - [x] Options: `--all`: traverse deletion boundaries; `--tree` show full tree vs only history; `--oneline` use a short diff option for each branch and print one line per branch. Maybe the printed info should be customizable with a template? Need to be careful with multi-line printing.
  - [x] Implement `--oneline` mode. This requires some sort of diffing mechanism, r.e. #212 (defer)
  - [x] Improve formatting style / appearance. Make sure to only write styles if printing to an interactive terminal
  - [x] A '--json' mode, which outputs a JSON representation of the full history, starting at the root, and then with nested entries for the children (or as a list, depending on whether `--all` is passed or not) (defer)
- [x] Allow accessing / reading past states by key (checking if valid, etc.), perhaps as hexadecimal
  - [x] Allow changing to a specific hexadecimal target with `autobib hist reset <key> <hex-target>`
  - [x] Use these states in the output of `autobib log`

### DIFF
--- a/src/app/cli.rs
+++ b/src/app/cli.rs
@@ -7,7 +7,7 @@ use std::{
 use anyhow::Result;
 use chrono::{DateTime, Local};
 use clap::{
-    CommandFactory, Parser, Subcommand, ValueEnum, builder::ArgPredicate, error::ErrorKind,
+    Args, CommandFactory, Parser, Subcommand, ValueEnum, builder::ArgPredicate, error::ErrorKind,
 };
 use clap_complete::aot::Shell;
 use clap_verbosity_flag::{Verbosity, WarnLevel};
@@ -207,9 +207,6 @@ pub enum Command {
         /// Set a field value using BibTeX field syntax
         #[arg(long, value_name = "FIELD_KEY={VALUE}")]
         set_field: Vec<SetFieldCommand>,
-        /// Update the modification time, inserting a new copy if the node has children.
-        #[arg(long)]
-        touch: bool,
     },
     /// Search for an identifier.
     ///
@@ -579,6 +576,17 @@ impl Command {
     }
 }
 
+#[derive(Debug, Args)]
+#[group(required = true, multiple = false)]
+pub struct IdTarget {
+    /// Apply to the record with this identifier.
+    #[arg(short, long)]
+    pub id: Option<RecordId>,
+    /// Apply to all records.
+    #[arg(short, long)]
+    pub all: bool,
+}
+
 /// Commands to manipulate version history.
 #[derive(Debug, Subcommand)]
 pub enum HistCommand {
@@ -610,15 +618,8 @@ pub enum HistCommand {
     Reset {
         /// The identifier for the reset operation.
         identifier: RecordId,
-        /// Set using a revision number.
-        #[arg(long, group = "reset_target")]
-        rev: Option<RevisionId>,
-        /// Set to the latest state with modification time preceding this date-time.
-        ///
-        /// This is a RFC3339 date-time, so make sure to indicate the timezone as well.
-        /// See, for example, the output of `autobib info -r modified`.
-        #[arg(long, group = "reset_target")]
-        before: Option<DateTime<Local>>,
+        /// The target active revision.
+        rev: RevisionId,
     },
     /// Insert new data for a deleted record, concealing any prior changes.
     ///
@@ -647,18 +648,22 @@ pub enum HistCommand {
         #[arg(long, value_name = "FIELD_KEY={VALUE}")]
         with_field: Vec<SetFieldCommand>,
     },
-    /// Move the database back in time.
+    /// Move record data back in time.
     ///
-    /// This is the same as calling `autobib hist reset --before` on every active entry in the database with
-    /// modification greater than the provided time.
+    /// Rewind specific records with `-i/--id`, or rewind all records with `-a/--all`.
     ///
     /// Use caution! The modification time may not correspond to the database state at the provided
     /// date-time if you have used `autobib hist (undo|redo|reset)`, since these methods only change the
     /// active state without introducing new changes. Your old data will still be retrievable, but
     /// it could require a lot of work to unwind the changes.
-    RewindAll {
+    Rewind {
         /// The datetime to rewind to.
+        ///
+        /// This is a RFC3339 date-time formatted like YYYY-MM-DD HH:MM:SS+HH:MM, with a trailing
+        /// timezone. See, for example, the output of `autobib info -r modified`.
         before: DateTime<Local>,
+        #[command(flatten)]
+        target: IdTarget,
     },
     /// Show all database changes in descending order by time.
     Show {
@@ -666,11 +671,13 @@ pub enum HistCommand {
         #[arg(long, value_name = "LIMIT")]
         limit: Option<u32>,
     },
-    /// Update the modification time of every active record in the database.
+    /// Update the modification time of records.
     ///
-    /// On success, print the new modification time of every active entry in
-    /// the database.
-    TouchAll,
+    /// On success, this prints the new modification time.
+    Touch {
+        #[command(flatten)]
+        target: IdTarget,
+    },
     /// Undo the most recent change associated with an identifier.
     Undo {
         /// The identifier for the undo operation.
--- a/src/app.rs
+++ b/src/app.rs
@@ -26,7 +26,7 @@ use etcetera::{AppStrategy, AppStrategyArgs, choose_app_strategy};
 
 use crate::{
     app::{
-        cli::{HistCommand, PruneCommand},
+        cli::{HistCommand, IdTarget, PruneCommand},
         log::print_log,
     },
     cite_search::{SourceFileType, get_citekeys},
@@ -296,7 +296,6 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
             update_entry_type,
             set_field,
             delete_field,
-            touch,
         } => {
             let cfg = config::load(&config_path, missing_ok)?;
             let nl = Normalization {
@@ -311,7 +310,7 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
                 delete_field,
             };
 
-            let no_non_interactive_cmd = nl.is_identity() && edit_cmd.is_identity() && !touch;
+            let no_non_interactive_cmd = nl.is_identity() && edit_cmd.is_identity();
 
             for key in identifiers {
                 let (Record { key, data, .. }, row) =
@@ -327,9 +326,7 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
                         // non-interactive command is requested, so we perform it without prompting
                         let mut editable_data = MutableEntryData::from_entry_data(&data);
 
-                        let mut changed = touch;
-                        changed |= editable_data.normalize(&nl);
-                        changed |= editable_data.edit(&edit_cmd);
+                        let changed = editable_data.normalize(&nl) || editable_data.edit(&edit_cmd);
 
                         if changed {
                             row.modify(&RawEntryData::from_entry_data(&editable_data))?
@@ -572,11 +569,7 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
                     None => {}
                 }
             }
-            HistCommand::Reset {
-                identifier,
-                rev,
-                before,
-            } => {
+            HistCommand::Reset { identifier, rev } => {
                 let cfg = config::load(&config_path, missing_ok)?;
                 if let Some((_, disambiguated)) = record_db
                     .state_from_record_id(identifier, &cfg.alias_transform)?
@@ -584,30 +577,24 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
                 {
                     let (_, state) = disambiguated.forget();
 
-                    if let Some(revision) = rev {
-                        match state.set_active(revision)? {
-                            RecordRowMoveResult::Updated(state) => {
-                                state.log_opt()?;
-                                state.commit()?;
-                            }
-                            RecordRowMoveResult::Unchanged(state, err) => {
-                                state.commit()?;
-                                match err {
-                                    SetActiveError::RowIdUndefined => {
-                                        error!("Revision does not exist in the 'Records' table");
-                                    }
-                                    SetActiveError::DifferentCanonical(remote_id) => {
-                                        error!(
-                                            "Revision exists, but it corresponds to a different record with canonical identifier '{remote_id}'"
-                                        );
-                                    }
+                    match state.set_active(rev)? {
+                        RecordRowMoveResult::Updated(state) => {
+                            state.log_opt()?;
+                            state.commit()?;
+                        }
+                        RecordRowMoveResult::Unchanged(state, err) => {
+                            state.commit()?;
+                            match err {
+                                SetActiveError::RowIdUndefined => {
+                                    error!("Revision does not exist in the 'Records' table");
+                                }
+                                SetActiveError::DifferentCanonical(remote_id) => {
+                                    error!(
+                                        "Revision exists, but it corresponds to a different record with canonical identifier '{remote_id}'"
+                                    );
                                 }
                             }
                         }
-                    } else if let Some(dt) = before {
-                        let state = state.rewind(dt)?;
-                        state.log_opt()?;
-                        state.commit()?;
                     }
                 }
             }
@@ -656,10 +643,28 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
                     None => {}
                 }
             }
-            HistCommand::RewindAll { before } => {
-                let snapshot = record_db.snapshot()?;
-                snapshot.rewind_all(before)?;
-                snapshot.commit()?;
+            HistCommand::Rewind {
+                before,
+                target: IdTarget { id, all },
+            } => {
+                if all {
+                    let snapshot = record_db.snapshot()?;
+                    snapshot.rewind_all(before)?;
+                    snapshot.commit()?;
+                } else if let Some(record_id) = id {
+                    let cfg = config::load(&config_path, missing_ok)?;
+                    if let Some((_, disambiguated)) = record_db
+                        .state_from_record_id(record_id, &cfg.alias_transform)?
+                        .require_record()?
+                    {
+                        let (_, state) = disambiguated.forget();
+                        let state = state.rewind(before)?;
+                        state.log_opt()?;
+                        state.commit()?;
+                    }
+                } else {
+                    unreachable!("ArgGroup requires one of these arguments");
+                }
             }
             HistCommand::Show { limit } => {
                 let snapshot = record_db.snapshot()?;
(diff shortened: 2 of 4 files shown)


alexrutar:
Nice looks like GitHub programming broke the macos runner.

https://github.com/actions/runner/issues/4134

https://github.com/orgs/community/discussions/180160

https://github.com/actions/runner-images/issues/13341

etc.

wupr:
Regarding a new PR -- one possibility is to start a temporary development branch and create PRs to that branch first, which would allow the flexibility of having incomplete code/docs and even doing some force-pushes.

Removing automatic alias creation sounds fair.

alexrutar:
Also see:

- https://github.com/autobib/autobib/blob/db-rework/ARCHITECTURE.md
- https://github.com/autobib/autobib/blob/db-rework/docs/data_model.md

docs/data_model.md
```diff
@@ -0,0 +1,113 @@
...
+
+1. `entry`: There is bibliographic data associated with the record.
+2. `deleted`: This is a deletion marker indicating that there is no data.
+   There may also be information about a replacement key.
+   This state is created with `autobib delete` or `autobib delete --replace`.
+3. `void`: This is a special state which is similar to there being no record present in the database.
+   For example, this will cause `autobib get` to automatically retrieve new data for the row.
+   It is uncommon for a record to be in this state, but this can be attained manually with `autobib hist void`, or is created automatically with `autobib hist rewind-all` or `autobib hist reset --before` when the threshold time precedes the existence of the record in the database.
+
+Note that `void` entries still contain some state: it preserve aliases, and still tracks the *prov-id* which can be used to more efficiently look-up new data.
```
wupr: "preserve" --> "preserves"

docs/data_model.md
```diff
@@ -0,0 +1,113 @@
...
+
+### Moving around the edit-tree
+
+The simplest operations regarding revisions are `autobib hist undo` and `autobib hist redo`.
+Undo sets the active state to the parent state, if the parent state exists.
+Redo sets the active state to the child if there is exactly one child; otherwise, it is required to pass an explicit index indicating which child to follow.
```
wupr: Is this still true?

docs/data_model.md
```diff
@@ -0,0 +1,113 @@
...
+
+For more complex operations involving the edit-tree, the output of `autobib log --tree` can be useful.
+This prints a branch diagram showing the relationship between all of the states ordered in reverse chronological order.
```
wupr: "ordered in reverse chronological order" --> "in reverse chronological order"

src/app/cli.rs
```diff
@@ -477,22 +558,152 @@ impl Command {
...
+        #[arg(long, value_name = "FIELD_KEY={VALUE}")]
+        with_field: Vec<SetFieldCommand>,
+    },
+    /// Move the database back in time.
+    ///
+    /// This is the same as calling `autobib reset --before` on every active entry in the database with
+    /// modification greater than the provided time.
+    ///
+    /// Use caution! The modification time may not correspond to the database state at the provided
+    /// date-time if you have used `autobib undo/redo/reset`, since these methods only change the
```
wupr: `autobib hist (undo|redo|reset)`

src/app/cli.rs
```diff
@@ -477,22 +558,152 @@ impl Command {
...
+        /// Set the entry type.
+        #[arg(long, value_name = "ENTRY_TYPE")]
+        with_entry_type: Option<EntryType>,
+        /// Set specific field values using BibTeX field syntax
+        #[arg(long, value_name = "FIELD_KEY={VALUE}")]
+        with_field: Vec<SetFieldCommand>,
+    },
+    /// Move the database back in time.
+    ///
+    /// This is the same as calling `autobib reset --before` on every active entry in the database with
```
wupr: `autobib hist reset --before`

src/app/cli.rs
```diff
@@ -164,57 +178,82 @@ pub enum Command {
...
     DefaultConfig,
     /// Delete records and associated keys.
     ///
-    /// Delete a record, and all referencing keys (such as aliases) which are associated with the
-    /// record. If there are multiple referencing keys, they will be listed so that you can confirm
-    /// deletion. This can be ignored with the `--force` option.
+    /// By default, this performs a soft delete, where the current data and keys are retained
+    /// but future attempts to read them will result in an error. The data can be recovered with
+    /// `autobib hist undo`. The key provided with the `--replace` option will be used to suggest
+    /// replacements.
```
wupr:
```suggestion
    /// replacements when the deleted data is requested.
```

src/app/cli.rs
```diff
@@ -164,57 +178,82 @@ pub enum Command {
...
-        force: bool,
+        /// The records to delete.
+        identifiers: Vec<RecordId>,
+        /// A replacement key.
+        #[arg(short, long, group = "delete_mode")]
+        replace: Option<RecordId>,
+        /// Hard deletion, which removes all history and aliases, and cannot be undone.
+        #[arg(long, group = "delete_mode")]
+        hard: bool,
+        /// Update aliases, either deleting or changing them to point to the new row if `--replace` is specified.
```
wupr:
```suggestion
        /// Update aliases, making them point to the new record if `--replace` is specified, or deleting them otherwise.
```

docs/changelog/next.md
```diff
@@ -1 +1,49 @@
...
+
+## Breaking changes
+
+- `autobib delete` now performs 'soft-deletion' by default, which does not remove the record from your database but instead inserts a deletion marker.
+  The old behaviour can be obtained with `autobib delete --hard`.
+  It is no longer necessary to pass all identifiers when performing a deletion.
+  Passing redundant identifiers will now result in an error.
+- `autobib merge` has been removed.
+  It is no longer possible to merge records together.
+  This functionality has been replaced by `autobib delete --replace <replacement>`, which performs a soft delete and stores a replacement identifier.
```
wupr: I don't think `autobib delete --replace <replacement>` completely replaces `autobib merge`, as the user can't easily merge even non-conflicting fields now. We should explain this.

alexrutar:
In the new version, merging is no longer possible - precisely for the reason that you have identified that when you merge, you lose information about aliases (so "unmerging" is impossible without some sort of procedure to decide where to point the aliases).

I was hoping that `autobib delete --replace` would be a sufficiently ergonomic fallback. Now instead of merging, you should run `autobib delete --replace` with the thing you would have merged into. Then, if you run `autobib get` with the "merged" record, in the current implementation it will fail but provide a warning indicating that it was merged into another record, and tell you what to use instead.

This behaviour could be updated to "auto-follow the replacement but warn"? Then `autobib merge` could be re-introduced, essentially as a wrapper around
```
autobib update <keep> --from-record <delete>
autobib delete <delete> --replace <keep>
```

The reason why I did not implement auto-follow was: (1) what about chain replacement (a -> b -> c) which makes following recursive; and (2) it seems weird to auto-follow if the target might not be in the database.

(Edit: maybe (2) could be implemented by having a special 'merged into' column, with a foreign key relation on the `name` column of the `Identifiers` table.)

I originally thought it would be nice for the replacement to point to a literal row in the 'Records' table - but there are foreign key issues here - and also what to do when updating rows?

Keen to hear your thoughts here since you seem to be using `merge` more than me!

(Edit 2: another funny foreign key trick which might work: point to the `record_key` column in 'Identifiers' and use `ON UPDATE CASCADE ON DELETE RESTRICT / SET NULL`; because 'Identifiers' is always updated this will correctly cascade the update.)

src/app/cli.rs
```diff
@@ -227,77 +241,144 @@ pub enum Command {
...
+    /// Replace an identifier with another one and merge the data.
+    ///
+    /// The original identifier must be present in the database. If the target identifier is not in
+    /// the database, its data will be retrieved first.
+    Replace {
+        /// The identifier to replace.
+        identifier: RecordId,
+        /// Replace with another identifier.
+        #[arg(short, long, group = "replace_target")]
+        with: Option<RecordId>,
```
wupr: We could Clap attribute `value_name = "IDENTIFIER"` for `identifier` and `with`

wupr (approved):
@alexrutar Happy new year! And very sorry about the long delay. I was taking the time to re-learn the code base but then got distracted by other things. I think I'll just put together some comments now instead of blocking this any further. Most functionalities I tried appeared to be working as expected. (The only thing I haven't tested is `autobib import`.)

#### Housekeeping

- `src/db/migrate/v1/records_new.sql` seems to be redundant (replaced by `src/db/migrate/v1/create_table_tmp_records.sql`).
- `tests/resources/dedup` needs to be renamed now since we no longer have `autobib dedup`.

#### Documentation

- Should we bring more attention to the change of terminology in the changelog?
- We are changing a lot of CLI options. We could mention these at one stroke and refer the user to the help messages, even if we don't enumerate all of them.
- In view of the upcoming docs website, it might make thing tidier if we moved `ARCHITECTURE.md` to `docs/architecture.md` and renamed `docs/data_model.md` to `docs/data-model.md`?

#### CLI

- How about making `autobib edit --touch abc` --> `autobib hist touch abc`, and `autobib hist touch-all` --> `autobib hist touch --all`?
- Similarly, how about `autobib hist reset --before time abc` --> `autobib hist rewind abc time`, and `autobib hist rewind-all` --> `autobib hist rewind --all`?
- ~~`autobib replace`: I thought the "target identifier" was the "current" one, and the "original identifier" was the "incoming" one, until I tried the command and noticed it was implemented the other way. And I can find arguments for both options. If the target is being remotely retrieved, then it's intuitively the incoming one. But it is the target record's history that has data written to it, so it seems like the target should be the current one.~~ I think you've changed this since I last read the code.

I have more notes, but they're less important and can be separate issues or PRs after this is merged.

src/app/cli.rs
```diff
@@ -169,52 +175,60 @@ pub enum Command {
...
-        citation_keys: Vec<RecordId>,
+        /// The record(s) to edit.
+        identifiers: Vec<RecordId>,
         /// Normalize whitespace.
         #[arg(long)]
         normalize_whitespace: bool,
         /// Set "eprint" and "eprinttype" BibTeX fields from provided fields.
-        #[arg(long, value_delimiter = ',', value_name = "FIELD_NAME")]
+        #[arg(long, value_delimiter = ',', value_name = "FIELD_KEY")]
         set_eprint: Vec<String>,
```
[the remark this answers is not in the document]
alexrutar: I've inlined these comments now instead of putting them at the top-level documentation! I think this is generally speaking a bit easier to read.

src/app/cli.rs
```diff
@@ -336,16 +414,16 @@ pub enum Command {
         /// Append new entries to the output.
         #[arg(short, long, requires = "out")]
         append: bool,
-        /// Retrieve records but do not output BibTeX or check the validity of citation keys.
+        /// Retrieve records but do not output BibTeX or check the validity of identifiers.
         #[arg(long, group = "output")]
         retrieve_only: bool,
-        /// Only print the citation keys which were found (sorted and deduplicated).
+        /// Only print the identifiers keys which were found (sorted and deduplicated).
```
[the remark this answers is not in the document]
alexrutar: In this specific case yes, since they're actual citation keys found in the file.

src/record/key.rs
```diff
@@ -340,15 +351,22 @@ impl RemoteId {
         full_id.push_str(alias.0.as_str());
         Self::new_unchecked(full_id, PROVIDER_LEN)
     }
+
+    pub fn get_ref(&self) -> RemoteId<&str> {
```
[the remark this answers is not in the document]
alexrutar: Hmm, I think `as_deref` isn't terrible in this case, except for the fact that there are no `T: Deref` bounds around currently. Perhaps I will rename to that for now (and it can be renamed later again if a better name shows up.)

src/app/cli.rs
```diff
@@ -227,99 +241,166 @@ pub enum Command {
...
     },
+    /// Manipulate version history.
+    Hist {
+        #[command(subcommand)]
+        hist_command: HistCommand,
+    },
     /// Import records from a BibTeX file.
+    ///
+    /// The implementation automatically determines a remote identifier from the data, using
+    /// your `preferred_providers` config setting and with unspecified fallback if there is no
```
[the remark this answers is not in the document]
alexrutar: I've fixed the phrasing!

alexrutar:
> @alexrutar Happy new year! And very sorry about the long delay. I was taking the time to re-learn the code base but then got distracted by other things. I think I'll just put together some comments now instead of blocking this any further. Most functionalities I tried appeared to be working as expected. (The only thing I haven't tested is `autobib import`.)
> 
> #### Housekeeping
> * `src/db/migrate/v1/records_new.sql` seems to be redundant (replaced by `src/db/migrate/v1/create_table_tmp_records.sql`).
> * `tests/resources/dedup` needs to be renamed now since we no longer have `autobib dedup`.
> 
> #### Documentation
> * Should we bring more attention to the change of terminology in the changelog?
> * We are changing a lot of CLI options. We could mention these at one stroke and refer the user to the help messages, even if we don't enumerate all of them.
> * In view of the upcoming docs website, it might make thing tidier if we moved `ARCHITECTURE.md` to `docs/architecture.md` and renamed `docs/data_model.md` to `docs/data-model.md`?
> 
> #### CLI
> * How about making `autobib edit --touch abc` --> `autobib hist touch abc`, and `autobib hist touch-all` --> `autobib hist touch --all`?
> * Similarly, how about `autobib hist reset --before time abc` --> `autobib hist rewind abc time`, and `autobib hist rewind-all` --> `autobib hist rewind --all`?
> * ~`autobib replace`: I thought the "target identifier" was the "current" one, and the "original identifier" was the "incoming" one, until I tried the command and noticed it was implemented the other way. And I can find arguments for both options. If the target is being remotely retrieved, then it's intuitively the incoming one. But it is the target record's history that has data written to it, so it seems like the target should be the current one.~ I think you've changed this since I last read the code.
> 
> I have more notes, but they're less important and can be separate issues or PRs after this is merged.

I've implemented this as `autobib hist touch --id abc` or `autobib hist rewind --id abc`! I think clap works a bit more cleanly with this type of syntax - plus it leaves flexibility in the future to allow multiple / combined targets (like, `--id abc --matches zbl:*` or similar).

PR #260, opened 2025-11-14
Release version 0.5.1

alexrutar:
A patch release would be nice to release the bug-fixes currently on `main`, and also to try out the new ARM bottles! Especially since #257 is a really large changeset, and will certainly be `0.6.0`.

See https://github.com/autobib/homebrew-autobib/pull/10#issuecomment-3530953730


wupr (approved):
OK! I've pushed some minor edits to the changelog. Maybe a squash-merge is easier.

Issue #265, opened 2025-11-29
Add `--oneline` and `--json` modes for `autobib log`


alexrutar:
JSON mode should print representation of the full history, starting at the root, and then with nested entries for the children (or as a list, depending on whether `--tree` is passed or not).

`--oneline` should print a short representation of the state (one per line). Also see #212 for probably related diff operations.

Issue #264, opened 2025-11-28
Improve implementation of `autobib info`


alexrutar:
1. If the output is piped, I think it should write JSON instead
2. Improve readability of `--report all`.

alexrutar:
Fixed in #329

Issue #158, opened 2025-01-01
Implement `autobib export`


alexrutar:
We could support exporting the database in various formats. Signature like `autobib export <path> [--force] [--format bib,sql]`

Few options for `--format sql`:

1. Use the [backup api](https://docs.rs/rusqlite-le/latest/rusqlite/backup/index.html) to create a byte-for-byte copy of the database.
2. Use [VACUUM INTO](https://www.sqlite.org/lang_vacuum.html) to make an equivalent but more optimized database.

This could also could be used to implement a 'batch' API for `autobib get`; but of course with the caveat that all operations are local only.

See also #104

alexrutar:
I'm cutting this into two separate issues: `autobib export` and `autobib backup`. I think `autobib export` should only write bibtex file, so it is really like the opposite of `autobib import` and is focused on generating records; sort of like a bulk version of `autobib get`. Then `autobib backup` contains `sql`, `JSON`, etc. output with the goal of creating an exact representation of the current database.

alexrutar:
R.e. https://github.com/autobib/autobib/issues/251#issuecomment-3511691064 it would also be reasonable to support non-bibtex output formats (CSV, JSON, etc.)

PR #268, opened 2026-01-03
Add support for `--multi` flag with `autobib find`

### LINKED ISSUE #248: Add support for `autobib find --multi`
This is actually an upstream issue in `nucleo-picker`, but I'm putting there here to keep track.

alexrutar:
This PR adds a new selection mode `autobib find --mode preferred-id` which, instead of returning the canonical ID, does a lookup and determines the preferred ID according to the configuration.

This PR allows adds support for multiple selections in `autobib find --mode canonical-id|preferred-id`.

Currently, multiple selections of attachments has not been implemented because the most ergonomic way to do this should be worked out. The naïve solution is to first allow multiple ID selections, and then for each ID allow multiple attachment selections; but I believe that this is confusing to use since one might lose track of which IDs the attachments belong to. Ideally, it would be possible to select them together, but I cannot think of an easy way to implement this. For this reason, I have decided to exclude it for now.

Should multiple selections be behind a flag, or always permitted? Currently, they are behind a flag.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -237,9 +237,9 @@ checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
 
 [[package]]
 name = "chrono"
-version = "0.4.43"
+version = "0.4.44"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fac4744fb15ae8337dc853fee7fb3f4e48c0fbaa23d0afe49c447b4fab126118"
+checksum = "c673075a2e0e5f4a1dde27ce9dee1ea4558c7ffe648f576438a20ca1d2acc4b0"
 dependencies = [
  "iana-time-zone",
  "js-sys",
@@ -906,9 +906,9 @@ checksum = "92ecc6618181def0457392ccd0ee51198e065e016d1d527a7ac1b6dc7c1f09d2"
 
 [[package]]
 name = "js-sys"
-version = "0.3.88"
+version = "0.3.91"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c7e709f3e3d22866f9c25b3aff01af289b18422cc8b4262fb19103ee80fe513d"
+checksum = "b49715b7073f385ba4bc528e5747d02e66cb39c6146efb66b781f131f0fb399c"
 dependencies = [
  "once_cell",
  "wasm-bindgen",
@@ -1279,9 +1279,9 @@ dependencies = [
 
 [[package]]
 name = "regex-syntax"
-version = "0.8.9"
+version = "0.8.10"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a96887878f22d7bad8a3b6dc5b7440e0ada9a245242924394987b21cf2210a4c"
+checksum = "dc897dd8d9e8bd1ed8cdad82b5966c3e0ecae09fb1907d58efaa013543185d0a"
 
 [[package]]
 name = "ring"
@@ -1372,9 +1372,9 @@ dependencies = [
 
 [[package]]
 name = "rustls"
-version = "0.23.36"
+version = "0.23.37"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "c665f33d38cea657d9614f766881e4d510e0eda4239891eea56b4cadcf01801b"
+checksum = "758025cb5fccfd3bc2fd74708fd4682be41d99e5dff73c377c0646c6012c73a4"
 dependencies = [
  "log",
  "once_cell",
@@ -1609,9 +1609,9 @@ dependencies = [
 
 [[package]]
 name = "tempfile"
-version = "3.25.0"
+version = "3.26.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0136791f7c95b1f6dd99f9cc786b91bb81c3800b639b3478e561ddb7be95e5f1"
+checksum = "82a72c767771b47409d2345987fda8628641887d5466101319899796367354a0"
 dependencies = [
  "fastrand",
  "getrandom 0.4.1",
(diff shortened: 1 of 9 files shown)


alexrutar:
@wupr Do you think this should be behind a flag? Should it be enabled / disabled by default? For example, `fzf` enables it by default. On the other hand, in the end I think this is not so compatible with attachments, and it would be weird to have the default change depending on the value for `--mode`.

alexrutar:
Having a flag also allows things like `--multi=N` which means "choose multiple items, but at most `N`" (and it is enforced within the picker interface itself)

(It is now behind an explicit flag.)

wupr:
I think multi-selection should be the default. One might only realise they need to make multiple selections after launching the picker and entering the search pattern. Not having to run the picker again (after either quitting or remembering the previous selection) is more user-friendly

alexrutar:
> I think multi-selection should be the default. One might only realise they need to make multiple selections after launching the picker and entering the search pattern. Not having to run the picker again (after either quitting or remembering the previous selection) is more user-friendly

Sounds good! In the most recent commit I did implement this, and then forgot to edit the comment. Also, `fzf` for example has multiple selections enabled by default.

wupr:
So I would advocate for a `--selection-limit` flag that takes a single-layer `Option`, and a `--single` flag instead of `--no-multi`.

alexrutar:
> So I would advocate for a `--selection-limit` flag that takes a single-layer `Option`, and a `--single` flag instead of `--no-multi`.

I agree this would be more precise; the current flags are to agree with `fzf`.  Is it worthwhile to use syntax that a user might not expect?

wupr:
Hmm I see. I don't know how `fzf` ended up with these flags, but I suspect default multi-select was an afterthought and they didn't want to break compatibility? I still think confusion with `fzf` is less likely than confusion due to bad naming (I didn't know about `fzf`'s `--multi` flag until just now despite being a daily user), so this is our opportunity to set the trend.

wupr:
> This PR adds a new selection mode `autobib find --mode preferred-id` which, instead of returning the canonical ID, does a lookup and determines the preferred ID according to the configuration.

I happen to have been thinking about output filters. The choice between `preferred-id` and `canonical-id` is one that is relevant for the output stage of `autobib find`, rather than the selection stage. Instead of adding a selection mode, we could generalise this new output option by allowing the user to supply a custom output template (same syntax as the picker rendering template) and adding a meta `%preferred_id`.

I realise `autobib find` is intended to be used with `autobib get` for getting record data. So alternatively we could extend `autobib get` (or `autobib info`) with output templates. But since the record data is already in memory during selection, maybe outputting directly would make sense too?

alexrutar:
> I happen to have been thinking about output filters. The choice between `preferred-id` and `canonical-id` is one that is relevant for the output stage of `autobib find`, rather than the selection stage. Instead of adding a selection mode, we could generalise this new output option by allowing the user to supply a custom output template (same syntax as the picker rendering template) and adding a meta `%preferred_id`.

I agree, this is much more sensible!

> I realise `autobib find` is intended to be used with `autobib get` for getting record data. So alternatively we could extend `autobib get` (or `autobib info`) with output templates. But since the record data is already in memory during selection, maybe outputting directly would make sense too?

Along these lines, I could also imagine adding a meta `%bibtex` which expands into the full record (like running `autobib get`). But now it is really starting to sound like `autobib get` should just have an output template? And the default template is just `{ %bibtex }` (also need to worry about separator between entries, we don't want a trailing newline).

Perhaps this can be done in a unified way.

alexrutar:
Hmm, on second thought, this is a bit weird - `autobib get` assumes its output is formatted as bibtex (for example, when using `--append`).

alexrutar:
On third thought, this could also simplify some other things too. Like, `autobib source` has a `--print-keys` option. This could just be part of the output template instead.

alexrutar:
A totally crazy thing which would complete the loop would be an 'input template'... unfortunately I think such a thing is impossible with the current template system but it is conceptually cool!

alexrutar:
Another thing to keep in mind: I think something like JSON format is a valid output type as well, so if we converge on output templates it needs to be able to handle this as well (outputting valid JSON is quite a bit more difficult than just having individual entry formatting + separator printing)

wupr:
> `autobib get` assumes its output is formatted as bibtex (for example, when using `--append`).

Ah I neglected this. `autobib get` should probably stay rigid then.

> autobib source has a --print-keys option. This could just be part of the output template instead.

I think `--print-keys` should be kept for debugging reasons. But I agree `autobib source` could allow output template as well.

> I think something like JSON format is a valid output type as well

JSON was one of the main things I was thinking of. I had to use `jq` in the Homebrew CI to process `gh` output. Being able to pipe autobib's output through `jq` would be very cool.

wupr:
JSON output was first proposed in https://github.com/autobib/autobib/issues/158#issuecomment-3511736675.

I guess #264 and #265 also have JSON-related proposals, but for metadata

alexrutar:
> I happen to have been thinking about output filters. The choice between `preferred-id` and `canonical-id` is one that is relevant for the output stage of `autobib find`, rather than the selection stage. Instead of adding a selection mode, we could generalise this new output option by allowing the user to supply a custom output template (same syntax as the picker rendering template) and adding a meta `%preferred_id`.

I still like this idea a lot, but there are technical problems. Most notably, it is not possible to determine `%preferred_id` using only the data inside the record itself: in order determine the preferred ID it is necessary to read the `CitationKeys` table to determine all identifiers which point to the given canonical ID, and to filter those according to configuration. In the current implementation, a template is essentially a pure function mapping record data to strings and does not have database read access (or config data, for that matter) in general.

alexrutar:
We could decide to pass all of this data to the template, but we would need to do it up-front and it would add decent overhead to every database lookup even if `%preferred_id` is not required.

It is not possible to store this data inside the record itself since it is config-dependent.

alexrutar:
Another option is to pull this out of the `autobib find` implementation and move it into `autobib info` instead. Since we need the extra db lookup, running `autobib info <canonical> -r preferred-id` is not so unreasonable in my opinion.

If we go down this route, `autobib info` should be re-worked to allow batch / STDIN input, and output in a more sane format

alexrutar:
> Another option is to pull this out of the `autobib find` implementation and move it into `autobib info` instead. Since we need the extra db lookup, running `autobib info <canonical> -r preferred-id` is not so unreasonable in my opinion.
> 
> If we go down this route, `autobib info` should be re-worked to allow batch / STDIN input, and output in a more sane format

@wupr I have now added `autobib info <id> -r preferred`. `autobib info` should be re-factored but I thought it is best to hold off until later.

alexrutar:
I have also implemented a command `autobib format`, which should play the role of an 'output filter'. But we could still add output formatting explicitly, since as you mentioned the data is already in memory.

alexrutar:
Failures are spurious (stack overflow errors?? and also, on a different test each time). Not sure what to do, other than wait.

PR #270, opened 2026-01-06
Add a terminology glossary for consistent naming

alexrutar:
This PR adds adds a terminology glossary for special names used by Autobib.

@wupr This is in bad shape currently, feel free to edit!

TODOs:

- [ ] Is `records` a good name? Can we think of better terminology to distinguish: (1) the bibliographic data itself; (2) the bibliographic data plus associated metadata which we might add later (which is tracked in the changelog); (3) the data which is constant over the entire edit tree.
- [ ] Is `edit-tree` a good name?
- [ ] Check that the terminology defined here is used consistently throughout the documentation elsewhere.
- [ ] Better word for "remote identifier"? How should we distinguish from a "local identifier"? Using "remote identifier" as an umbrella term to include "local identifier" is definitely confusing (as was previously discussed somewhere).

### DIFF
--- a/docs/terminology.md
+++ b/docs/terminology.md
@@ -1,0 +1,21 @@
+# Glossary of terminology
+
+## Data terminology
+
+- *record*: a snapshot of bibliographic data along with associated metadata
+- *record data*: the bibliographic data associated with a record
+- *revision*: a hexadecimal string in one-to-one correspondence with records
+- *edit-tree*: the tree of records associated with a canonical identifier containing the history
+- *active record*: the unique record in the edit-tree which is currently accessible by an identifier
+
+## Identifier terminology
+
+- *identifier*: a text string referring to an active record
+- *canonical identifier*: a special identifier uniquely associated with the active record
+- *reference identifier*: additional identifiers which refer to a canonical identifier
+- *alias*: a custom identifier not containing a colon `:`
+- *provenance*: the specific origin of record data
+- *provider*: a named source from which record data can be obtained; the part before the `:` in a canonical or reference identifier.
+- *sub-id*: a special identifier used by a provider; the part after the `:` in a canonical or reference identifier
+- *local identifier*: a special canonical identifier for data without a named provider
+- *remote identifier*: a canonical or reference identifier referring to data which originates from a named provider


wupr:
To partially address the questions above:

1. I think (1) could be the "record content", and (2) is the "content and metadata", which we might casually refer to as the "record data", maybe we don't need to define it strictly. I'm not entirely sure now what you meant with (3).
2. I think "edit-tree" is a very intuitive name.

I want to pose a couple of questions as well:

5. Is a "revision" just the hexadecimal string, or is it also the record at that revision (which could be verbally convenient)?
6. The canonical ids are in one-to-one correspondence with both active records and edit-trees. Which correspondence should we think of as the principal one? (i.e. are canonical ids in correspondence with edit-trees because they are in correspondence with active records, or vice versa?) Is there a term for the collection of all records in an edit-tree?

wupr:
What's your opinion on using LLMs for the non-creative, non-decision-making parts of putting together a glossary? I think this is the kind of work they would be good at

alexrutar:
> What's your opinion on using LLMs for the non-creative, non-decision-making parts of putting together a glossary? I think this is the kind of work they would be good at

In this case I would think I disagree, mainly because I think the value of a glossary is only 50% the existence of the glossary itself; the other half of the value is actually going through the process of putting one together in order to land on good mental model for talking about the program itself!

On the other hand I think LLMs would be really good for:

1. Checking consistency of the glossary text with the way the text is used elsewhere in docs or help messages
2. Checking completeness of the glossary / identifying things that could be defined.

wupr:
> the other half of the value is actually going through the process of putting one together in order to land on good mental model for talking about the program itself!

I think we actually agree. I attempted to qualify with "the non-creative, non-decision-making parts", but I should clarify what I meant. The existing doc files already "imply" a mental model, which may or may not be consistent or an accurate reflection of how we were thinking of them. An LLM could make the implied mental model explicit, which we can check to compare with things in our minds and identify decisions we need to make, which is a creative part of the process.

alexrutar:
I see, then yes we are on the same page! I've been rather busy for the past few weeks (and will continue to be for a few more weeks) but will have more time after.

alexrutar:
Closing in favour of #329

PR #274, opened 2026-04-16
Release v0.6.0

alexrutar:
New release for `0.6.0`. I've been running `main` for quite a while now and have not run into any serious issues, so I think it is a good time for a new release.

@wupr


wupr:
Sounds good.

Do we want to do a dependency bump? `cargo audit` shows several vulnerabilities, including https://rustsec.org/advisories/RUSTSEC-2026-0099

alexrutar:
> Sounds good.
> 
> Do we want to do a dependency bump? `cargo audit` shows several vulnerabilities, including https://rustsec.org/advisories/RUSTSEC-2026-0099

Good idea, I've updated to rust 1.94 as well (except I did not update the `bincode` since this will require more work and there seem to be no issues with it, except that it is currently unmaintained).

PR #281, opened 2026-05-03
Fix zbmath

### LINKED ISSUE #279: zbmath id assume they are at least 7 digits (not correct)
For instance:
```
$ autobib -v get zbl:1044.11582
info: Autobib version: 0.6.0 (database version: 2)
info: SQLite version: 3.53.0
info: Using default database file '/home/tornaria/.local/share/autobib/records.db'
info: On-disk database version: 2
info: Interactive: true
info: Read-only: false
info: Configuration file not found at path '/home/tornaria/.config/autobib/config.toml'; using default configuration
info: Resolving remote record for zbl:1044.11582
error: Provider error: Reference source returned an invalid identifier: 'zbmath:814533'
```
and
```
$ autobib get zbmath:814533
error: Invalid key 'zbmath:814533': sub-id is invalid for the given provider
       Maybe you meant: 'mr:0814533'
```

Something as simple as:
```
--- a/src/provider/zbmath.rs
+++ b/src/provider/zbmath.rs
@@ -5,13 +5,8 @@ use super::{BodyBytes, Client, MutableEntryData, ProviderError, StatusCode, Vali
 use self::response::Response;
 
 pub fn is_valid_id(id: &str) -> ValidationOutcome {
-    if id.len() == 8 && id.as_bytes().iter().all(u8::is_ascii_digit) {
+    if id.len() <= 8 && id.as_bytes().iter().all(u8::is_ascii_digit) {
         ValidationOutcome::Valid
-    } else if id.len() == 7 && id.as_bytes().iter().all(u8::is_ascii_digit) {
-        let mut normalized = String::with_capacity(8);
-        normalized.push('0');
-        normalized.push_str(id);
-        ValidationOutcome::Normalize(normalized)
     } else {
         ValidationOutcome::Invalid
     }
```
seems to fix the issue, although I'm not sure what is the role of the `ValidationOutcome::Normalize(..)` result. If important, it should pad to 8 digits any number (not only 7 digits).

### LINKED ISSUE #280: Missing year
This is missing year:
```
$ autobib get zbmath:16716
@incollection{zbmath:16716,
  author = {Birch, B. J.},
  language = {English},
  pages = {191--212},
  title = {Hecke actions on classes of ternary quadratic forms},
  zbl = {0748.11023},
  zbmath = {00016716},
}
```
Note that the year is there in the api:
```
$ curl -s https://api.zbmath.org/v1/document/16716 | jq -r .result.year
1991
```

alexrutar:
Fixes the zbmath issues pointed out in #279 and #280

I wonder if 0-padding to 8 digits is correct. Maybe the normalization should instead strip leading zeros instead? Unfortunately this requires database version bump (to correctly migrate the previous `zbmath:` identifiers) but might be a good idea in case there are eventually 9 digits, somehow.

### DIFF
--- a/src/format.rs
+++ b/src/format.rs
@@ -256,20 +256,20 @@ impl Template {
         let mut ctx = init();
         for span in self.template.spans() {
             match span {
-                Span::Expr(Expression::Bare(Atom::FieldKey(k))) => {
-                    if !contains(k.as_ref(), &mut ctx) {
-                        return false;
-                    }
+                Span::Expr(Expression::Bare(Atom::FieldKey(k)))
+                    if !contains(k.as_ref(), &mut ctx) =>
+                {
+                    return false;
                 }
-                Span::Expr(Expression::IfDefined(k1, Atom::FieldKey(k2))) => {
-                    if contains(k1.as_ref(), &mut ctx) && !contains(k2.as_ref(), &mut ctx) {
-                        return false;
-                    }
+                Span::Expr(Expression::IfDefined(k1, Atom::FieldKey(k2)))
+                    if contains(k1.as_ref(), &mut ctx) && !contains(k2.as_ref(), &mut ctx) =>
+                {
+                    return false;
                 }
-                Span::Expr(Expression::IfUndefined(k1, Atom::FieldKey(k2))) => {
-                    if !contains(k1.as_ref(), &mut ctx) && !contains(k2.as_ref(), &mut ctx) {
-                        return false;
-                    }
+                Span::Expr(Expression::IfUndefined(k1, Atom::FieldKey(k2)))
+                    if !contains(k1.as_ref(), &mut ctx) && !contains(k2.as_ref(), &mut ctx) =>
+                {
+                    return false;
                 }
                 _ => {}
             }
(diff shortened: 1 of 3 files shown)


tornaria:
zbmath.org itself produces this bibtex:
```
@misc{zbMATH00016716,
 author = {Birch, B. J.},
 title = {Hecke actions on classes of ternary quadratic forms},
 year = {1991},
 language = {English},
 howpublished = {Computational number theory, {Proc}. {Colloq}., {Debrecen}/{Hung}. 1989, 191-212 (1991).},
 keywords = {11E45,11E20,11F11,11Y35,11G05,11F37,11F27},
 zbMATH = {16716},
 Zbl = {0748.11023}
}
```
so, padding to 8 digits for the label, but no padding in the bibtex entry.

When matching ids, they allow and ignore any number of leading zeros, e.g. https://zbmath.org/000000016716

wupr:
> but might be a good idea in case there are eventually 9 digits, somehow.

I think this is a convincing argument.

If we were to change the direction of normalisation, we could wait until the next time a version bump is required for other reasons.

Issue #241, opened 2025-10-27
Local configuration file?


alexrutar:
It would be nice to allow project-specific configuration. A few possibilities:

1. Support an optional local configuration file, say in `.autobib/config.toml`.
2. Allow manually passing config values with an option like `--config-str`. The syntax would follow [toml inline tables](https://toml.io/en/v1.0.0#inline-table) and each value would override whatever exists in the loaded configuration or the config file specified with `--config`.

Maybe I lean towards 2. more, since I personally find inheritance / priority logic to be unnecessarily confusing, and also requires us to implement 'merge' logic. (2) encourages only patching commands when really necessary, and encourages creating a new config file and passing it as an explicit argument with `--config` if necessary.

@wupr Curious what you think here

PR #282, opened 2026-05-04
Update zbmath internal format

### LINKED ISSUE #191: Implement automatic cleanup code
Clean up:

- Attachments directory
  - Use `WalkDir` to list all empty directories
  - Use `max_depth` (https://docs.rs/walkdir/latest/walkdir/struct.WalkDir.html#method.max_depth) to only delete empty directories that we automatically create (the 'path' hash components)
  - Use `contents_first` to iterate depth-first, and build a `Set` consisting of the directories which are known to be empty, so if a directory is non-empty but only contains an empty directory we can add it to the 'deletion' set
- Some sort of backup file directory (maybe at `~/.local/share/autobib/backups`)

### LINKED ISSUE #231: Implement `autobib backup`
We should run `autobib backup` to run one of the following:

1. Use the [backup API](https://docs.rs/rusqlite/latest/rusqlite/backup/struct.Backup.html) to backup the database to a new location.
2. [`VACUUM INTO ...`](https://www.sqlite.org/lang_vacuum.html)
3. Export a JSON file which a relatively plain machine-readable format (say, records + citation keys, but no changelog).

### LINKED ISSUE #273: Detect zbMath 'conflicting license' issues
I don't know of a good solution for this but here is the issue: in some cases zbMath does not return bibliographic data because of 'conflicting licenses'. For example, `autobib get zbl:1472.11214` returns
```bib
@article{zbl:1472.11214,
  author = {zbMATH Open Web Interface contents unavailable due to conflicting licenses. and zbMATH Open Web Interface contents unavailable due to conflicting licenses.},
  doi = {10.1080/00029890.2019.1651166},
  eprint = {1472.11214},
  eprinttype = {zbl},
  issue = {10},
  journal = {Am. Math. Mon.},
  language = {zbMATH Open Web Interface contents unavailable due to conflicting licenses.},
  origtitle = {zbMATH Open Web Interface contents unavailable due to conflicting licenses.},
  pages = {zbMATH Open Web Interface contents unavailable due to conflicting licenses.},
  publisher = {Mathematical Association of America (MAA), Washington, D.C.; Taylor \ Francis, Abingdon, Oxfordshire},
  title = {zbMATH Open Web Interface contents unavailable due to conflicting licenses.},
  volume = {126},
  year = {2019},
  zbl = {1472.11214},
  zbmath = {07135424},
}
```
The issue is that all of the fields are present so there is no error; they are just replaced with the text `zbMATH Open Web Interface contents unavailable due to conflicting licenses.`

The most reasonable behaviour I can think of here is to think of a good mechanism for providers to emit warnings, and to automatically prune the 'conflicting licenses' fields and just return
```bib
@article{zbl:1472.11214,
  doi = {10.1080/00029890.2019.1651166},
  eprint = {1472.11214},
  eprinttype = {zbl},
  issue = {10},
  journal = {Am. Math. Mon.},
  publisher = {Mathematical Association of America (MAA), Washington, D.C.; Taylor \ Francis, Abingdon, Oxfordshire},
  volume = {126},
  year = {2019},
  zbl = {1472.11214},
  zbmath = {07135424},
}
```

### LINKED ISSUE #290: Handle attachments with `autobib replace` and `autobib delete`
Currently, `autobib replace` and `autobib delete` do not merge attachments.

1. Should we provide a dedicated command for merging attachment directories?
2. What is the correct behaviour on replace / delete? (a) do nothing, (b) warn on orphaned, or (c) merge automatically?

The main problem with merging attachment directories is as far as I am aware there is no good way to avoid clobbering existing directories on rename. (You can check existence beforehand, but it is a TOCTOU error of course.) This is avoidable on common platforms by adding a dependency (e.g. https://github.com/indianakernick/renamore/tree/master/src) but that's somewhat less ideal.

alexrutar:
(Note: I tagged on a decent number of extra changes to this PR not in this description, see the changelog file edits for a summary)

Updates the internal zbmath format to remove 0-padding and migrates attachment format. Justification for removing 0-padding:

1. This is what is internally used by zbMath
2. More canonical (e.g., still works if identifiers have 9 numbers at some point)
3. More compact

Also updates the attachment folder format to use a format file `.autobib-format`. The folder format is determined by the contents of that file:

1. empty: legacy format (0-padded zbmath identifiers, `rapidhash::v1`) and base-32 with padding
2. `v1-migrating`: a mix of `v0` and `v1` format
3. `v1`: new format (`rapidhash::v3`) and base-32 without padding

In order to synchronize access to the attachment directory, this PR uses filesystem locks on the `.autobib-format` file. When not migrating, we acquire a shared lock, and in order to migrate we acquire an exclusive lock. This prevents migration from happening simultaneously to another Autobib process reading / writing the attachments directory.

This version does not migrate by default, but does provide a migration function `autobib util migrate-attachments` which converts from the `v0` format to the `v1` format. This version also is able to read and write both the `v0` and `v1` formats. Also, we now return an error if the attachment format is unknown (all of `v0`, `v1-migrating`, and `v1` are missing); this makes it safer to create `v2` in case we ever need to do this.

Future plan (subject to change):

1. In `autobib v0.8.0`, warn about `v0`directories  and suggest manual migration. Warn that migration will be automatic in a later version.
2. In `autobib v0.9.0`, automatically migrate
3. In some future version remove the format code and commit to a fixed attachment format.

Unfortunately, this PR is quite complicated and all for code that will (eventually) be deleted, I guess by `autobib v1.0`. In the long run, I think we should not require a folder format at all, and just fix a format and stick with it for perpetuity even if it requires jumping through more hoops to maintain backwards compatibility in the future.

I think it is not worth supporting alternative directory formats. In my opinion, this introduces extra complexity and decreases portability since it makes it harder for external programs to read the format.

Migration of the attachment directory is performed without reference to the database, and instead works by walking the attachment directory. For each directory:

1. Decode the existing identifier (ignoring the hash)
2. Compute the hash, the (expected) old directory path, and the new directory path
3. Compare the old and new paths; if equal to new, do nothing; if equal to old, rename; otherwise, warn on unrecognized format.

Directories in an unexpected format are logged, but not migrated.

If for some reason the rename target exists, this is reported, and the user is asked to manually merge the paths. Afterwards, the migration can be resumed / completed by re-running `autobib util migrate-attachments`.

In practice, migration issues should be very rare. Since the database version also changes, in most cases trying to open a migrated attachment directory with an old version of `autobib` will result in a version conflict with the database before even trying to read the attachments folder. The only way to get into trouble is to:

1. Have multiple databases, with different versions, say `a.db` and `b.db`
2. Migrate an attachment directory from `v0` to `v1`, while opening `a.db`.
3. Use an old version of `autobib` with database `b.db`, but with the migrated attachment directory.

And even here, the only problem will be that the attachments will not show up.


alexrutar:
@wupr I have a surprisingly large dilemma about the attachment directory situation and I am curious for your opinion on this. How should the 'version' state for the attachments directory be stored?

1. Implicitly using the database version. On one hand this is not good, since I think of the attachment directory as independent of the database. It should be possible for multiple databases to refer to the same attachment set. Also the attachment directory can be set independently of the database path. But maybe this is not a good way to think about it. Perhaps we should think of `(attachments, records.db)` as a pair to be migrated together at all points.
2. Explicitly using a file `$AUTOBIB_ATTACHMENTS_DIRECTORY/format.toml`. This could contain a `version = ...` field, along with any other state we would like to store. (Or perhaps `$AUTOBIB_ATTACHMENTS_PATH/state.db`.)
3. Inside `config.toml`. An upside is that this puts the user very transparently in control of what happens inside the attachments folder. E.g., have a section `[attachments]`; if `version = 0` (or `version` is unset) use `0`-padding for zbmath identifiers; etc. (could also allow configuration for other attachment path formats, etc.)

1,3 have the downside that they make the attachments directory entirely non-portable. But is this level of portability something we really want to preserve? Perhaps being non-portable is okay?

2 has the downside that it makes the application state distributed over even more area (now we have to keep track of `config.toml`, `records.db`, and `attachments/format.toml`!)

Something to keep in mind is that there are two types of state: (1) user configuration, and (2) information about the current state of the attachments directory. This is relevant for both 2 and 3. For 1 it is fine since user config stays in `config.toml` and other state can be stored inside `records.db`.

Finally, we could abandon this PR and keep zbmath 0-padded and not worry about this at all.

Maybe there is a better solution that I am failing to think of entirely.

alexrutar:
Another option is to just break very non-portably now: don't try to store state (i.e. keep doing what we are doing now); then hope we don't have to break the `attachments` directory format in a non-portable way in the future (probably wishful thinking); and use heuristics in the code when migrating (e.g., migrate explicitly at database change, check both current and legacy paths for some time afterwards, etc.).

alexrutar:
I checked what e.g. calibre does, and it looks metadata and attachments are stored together. At the same time, the file format does look fixed, and also human readable.

wupr:
@alexrutar I'd prefer portability to be preserved, so among the three, I'd vote for option 2.

Would a flag like `legacy_zbmath_padding` be better than a simple numeric version? If the flag is on, then we check both locations; otherwise, only the new location. At least right now I don't see a strong argument for linear versioning. We'd also have some more flexibility on how long we want to keep the flag on by default (e.g. until Autobib version 1.0). And future migrations could be done independently of zbMATH id padding.

Actually, if we use a flag, then maybe option 3 is also fine, with minimal impact on portability.

alexrutar:
I guess the config option has the following issue: how do we keep the `config.toml` state in sync with the attachments folder? Without storing some state inside the attachment folder itself, we resort to using heuristics to guess the format. And then this is essentially equivalent to just making a breaking change at some point in the future. Either that or we rely on the user to simultaneously change their config and manually migrate the format at the same time (seems annoying).

A possibility is to just break compatibility between versions but still keep portability (by having a unique file format):

1. In 0.7, don't migrate the attachments format: modify the binary to be able to read both legacy and new formats, but still only write the legacy format. (
2. Once adoption of binaries >= 0.7 is widespread, start writing in the new format and migrating the old format. Leave a marker file at the root of the attachments directory to indicate if it has been migrated or not. Then the 0.7 binaries can read the new format written by the 0.8 binaries.

Another thing that bothers me a bit is the hash format, since rapidhash changed their v1 format. (My fault for choosing a hash format that was too new...) The v3 format seems more stable. It would be nice to use it instead, or we can just use sha256 or something slower but totally stable. If we change the hash format at the same time it becomes trivial to detect the attachment directory format even without a marker file: read an attachment dir, decode the ID, compute the old hash and the new hash, and check which one agrees with the root.

alexrutar:
I think I agree 2 is the best.

The main thing is to guarantee forward (in)compatibility to make future changes easier if needed. Plan:

In 0.7.0: do not change the identifier format, but start checking that a file "format.toml" (name to be decided) is at the root of the attachments directory. It permits keys:

- `legacy_zbmath_id_format`: bool, default true
- `hash_strategy`: enum `rapidhash_v1`, `rapidhash_v3`, `sha256`, `none`, default rapidhash_v1 (`none` stores flat, without the hash prefix)

Read/write files according to the format described in the file. If there are any unknown keys or if there is an unrecognized value then return an error (since it indicates that a future version has changed the format). To be compatible with < 0.7.0, do not change the format, and write the current defaults if the file is missing.

In 0.8.0: Provide a new `autobib util` subcommand to change the hash strategy and id format. Warn on legacy_zbmath_id_format = false, and prompt to migrate (?)

In 0.9.0: migrate id format automatically (?)

etc.

alexrutar:
Also maybe it is better if don't let the user specify the format? I.e. just choose the migration target and provide a single command "autobib util update-attachments". And eventually (say, at 1.0) commit to making the format backwards compatible eternally, and stop reading the format file.

I'm a bit worried about doing updates atomically, which is why I think maybe we should make it hard to change the format. (Databases really are great.)

I just remember there was a discussion once to allow flat attachment storage. Not sure if we want to support it.

wupr:
Oh I neglected the fact that older versions would not even recognise the new flags / attachments format. Good point.

The three-step plan makes sense. In the second step (0.8.0 or possibly later), we'd want a good way of nudging users towards migration without getting in their way too much.

> just choose the migration target and provide a single command "autobib util update-attachments"

We could also forbid certain paths -- `legacy_zbmath_id_format` can only be turned off, and `hash_strategy` cannot go from `rapidhash_v3` to `rapidhash_v1`.

> a file "format.toml" (name to be decided)

Maybe `.autobib-format`.

wupr:
> 2\. Once adoption of binaries >= 0.7 is widespread

This reminds me that Homebrew doesn't collect telemetry for custom taps, although we do have the stats on crates.io.

alexrutar:
I realize there's an issue with atomicity. In principle the attachments directory could be partially updated (e.g. if it's slow and the user hits ctrl-c in the middle).

This would be a benefit of granular updates. Migrate each directory individually and (for a few versions) check both the old and new locations every time. Before migration starts create canary file (like '.version/1') and check for existence of a file like '.version/n' where n is too large for the current version before reading. It's not perfectly atomic but quite close, mainly it handles fragmented updates. Or something more sophisticated than this.

I really don't like re-implementing databases manually...

alexrutar:
Network failures are spurious, will try again later.

alexrutar:
@wupr I edited the comment at the top to summarize what I actually implemented.

alexrutar:
I changed the implementation to directly use `fs::rename` since it turns out that when renaming a folder, the target will never be overwritten unless it is an empty folder.

src/app/attach/migrate.rs
```diff
@@ -0,0 +1,230 @@
...
+    let migrating = format_root.join(FORMAT_V1_MIGRATING);
+    let v1 = format_root.join(FORMAT_V1);
+
+    match fs::remove_dir(&v0) {
+        Ok(()) => {
+            // `v0` exists and we just deleted it, so create `v1-migrating` to indicate that
+            // migration is in progress
+            fs::create_dir_all(&migrating)?;
+            Ok(())
+        }
```
wupr: If the program gets interrupted between removing the v0 marker and creating the migrating marker, then the user will get stuck as `.autobib-format` is empty?

This is a very small gap, but maybe we should rename v0 to migrating instead of removing and creating?

src/app/attach/migrate.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    match fs::read_dir(source) {
+        Ok(_) => {}
+        Err(err) if err.kind() == io::ErrorKind::NotFound => return Ok(true),
+        Err(err) if err.kind() == io::ErrorKind::NotADirectory => {
+            warn!(
+                "Skipping invalid path which is not a directory: {}",
+                source.display()
+            );
+        }
```
wupr, line 171: This case is probably missing a return

src/app.rs
```diff
@@ -385,11 +393,13 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
 
             match find_mode {
                 FindMode::Attachments => {
+                    let attachment_root =
+                        get_attachment_root(&data_dir, cli.attachments_dir, cli.read_only)?;
```
wupr, line 480: Should we force `true` for `read_only` here, like the behaviour of `autobib path` (:957)?

src/app/cli.rs
```diff
@@ -740,4 +744,12 @@ pub enum UtilCommand {
...
         deleted: bool,
     },
+    /// Migrate attachment directories to the current format.
+    MigrateAttachments,
+    /// Cleanup the attachment directory.
+    CleanupAttachments {
+        /// Delete empty subdirectories
+        #[arg(short, long)]
+        empty: bool,
+    },
```
wupr: Should we mention that currently if you don't pass any flag, this subcommand doesn't do anything?

src/app/import.rs
```diff
@@ -283,45 +277,61 @@ fn create_alias_and_commit(
...
-    canonical.extend_attachments_path(target_path);
-    match source_path.file_name() {
-        None => anyhow::bail!("Cannot import filename containing relative path"),
-        Some(file_name) => {
-            fs::create_dir_all(&target_path)?;
-            target_path.push(file_name);
-            // FIXME: this is a TOCTOU error
-            if !target_path.exists() {
-                fs::copy(source_path, target_path)?;
+/// File import struct with stratch space to reduce re-allocation.
```
wupr: Typo: "stratch" -> "scratch"

src/path_hash.rs
```diff
@@ -1,43 +1,197 @@
...
+
+    /// Overwrite the provided buffer with the attachment directory corresponding to the identifier.
+    pub fn attachment_dir_in(&self, id: &RemoteId, path: &mut PathBuf) {
+        path.clear();
+        path.push(&self.root);
+        match self.format {
+            AttachmentFormat::V0 => RemoteIdAttachmentPathV0(id).extend_attachments_path(path),
+            AttachmentFormat::V1 => RemoteIdAttachmentPathV1(id).extend_attachments_path(path),
+        }
+    }
```
wupr, line 275: It's worth adding a comment here that this method is to reduce allocation. I initially thought this method was unnecessary, as `path` can always be updated with an assignment

docs/changelog/next.md
```diff
@@ -1,9 +1,25 @@
...
+## Breaking changes
 
+- `zbmath` identifiers are now stored internally without 0-padding to length 8
 - SQLite is now only bundled when the Cargo feature `bundled-sqlite` is enabled.
   This feature is enabled by default, but this may cause breakage with builds using `--no-default-features`.
   Disabling this feature will cause the compiled binary to link to your SQLite system library instead.
+
+## Changes
+
+- Autobib is migrating to a new attachment folder format.
```
wupr: We should stress that migration is not automatic with this release

docs/changelog/next.md
```diff
@@ -1,9 +1,25 @@
...
+  The new folder format is not compatibile with autobib versions `< v0.7.0`.
```
wupr: `< 0.7.0`. Since you already said "versions", the `v` is not necessary. (`autobib --version` also doesn't print the `v`.)

wupr (approved):
Sorry for the wait. The changes look nice and sensible, and I think you've covered most edge cases (one comment on an edge case below).

> Instead of using a file, this PR uses the presence of directories since mkdir and rmdir are atomic operations. The process isn't perfectly threadsafe, but it's basically as close as possible while being platform-independent and should be fine in practice.

I would have suggested creating an empty file `v0`/`v1-migrating`/`v1` inside `.autobib-format/`, and rename it as necessary. I think creating and renaming files can be atomic? Didn't look too carefully into this. Directories as they are implemented are OK too, but I (and I imagine some others) have a habit of digging into config folders and cleaning up empty directories... but then that's probably rare and once we fix the format, it's irrelevant anyway.

src/app.rs
alexrutar: Good catch! Yes, I just missed it when I changed `autobib path`.

src/app/attach/migrate.rs
alexrutar: The problem is that rename will succeed even if `v1-migrating` exists, since it is empty. But maybe there is a way around this by making `v0` etc. non-empty? Like `.autobib-format/v0/CANARY`. Then it both resolves your 'empty directory' issue and also this renaming issue. I will think about it!

alexrutar:
Renaming files without overwriting is a bit of a platform-specific mess, unfortunately, at least based on what I read. On linux and macos there are specific syscalls (one can use `rustix` to abstract over this), but on Windows it is different, and on other platforms (most notably `(Free|Open)BSD`) it is potentially not even supported...

The most reliable 'rename without overwriting' is to rename a non-empty directory to a non-empty directory which as far as I can tell will not overwrite the target on the majority of platforms (at least the ones above).

src/app/attach/migrate.rs
alexrutar: On second thought, if `v0` and `v1-migrating` both exist I think this should be considered that the state is `v0`. Perhaps using `rename` is fine. Need to think if this has any impact elsewhere...

alexrutar:
@wupr I made the directory operations more atomic in general. I noticed one extra race condition:

- Thread A sees `v0`
- Thread B migrates from `v0` to `v1`
- Thread A creates a new attachment in `v0` format

How worried should we be about this?

alexrutar:
Three options off the top of my head:

Easy fix is to ignore or provide a 'retry' option for migration since the state created by adding a `v0` attachment into a `v1` directory is equivalent to an interrupted migration.

Intermediate fix is to check that attachment dir format is the same before and after adding the attachment, and if it is not return an error explaining the above.

The hard fix is to use a lockfile (lots of extra rather complex code).

wupr:
Ah that feels like a comparatively likely scenario, especially with network storage. If a lock file is too complex, I think the intermediate fix is good enough.

alexrutar:
> Ah that feels like a comparatively likely scenario, especially with network storage. If a lock file is too complex, I think the intermediate fix is good enough.

Hmm, we could also try to make `autobib attach` more atomic.

alexrutar:
I managed to get an epsilon of cell connection! I re-wrote the implementation to use a lockfile implementation. Contrary to what I said before, it is about as complicated as the previous code.

Currently uses `$AUTOBIB_ATTACHMENTS_DIRECTORT/LOCKDIR`. I thought it should be not hidden and obviously named, so that other applications can write / respect the directory as well.

In read-only mode, we only check for `LOCKDIR` at startup and abort if it exists. This means that reads could become invalid if writes happen while reading. I hesitate to add a `--locked` option, because eventually we will remove the lockdir implementation and only perform atomic operations in the attachments directory.

In read-write mode, we create `LOCKDIR` at startup and remove it when closing.

alexrutar:
~~The only extra lockfile complexity is signal handling; `drop` does not run on `CTRL-C`.

https://detegr.github.io/doc/ctrlc/

There is a method to delete spurious lockfiles `autobib util cleanup --lockdir`, but it is so common for someone to use `CTRL-C` to quit things that maybe we should handle this automatically?~~

alexrutar:
~~Ok, everything is up to date: the only unresolved question on my list is how hard we should try to handle lockfiles when `ctrl-c` which prevents cleanup.~~

alexrutar:
> Ok, everything is up to date: the only unresolved question on my list is how hard we should try to handle lockfiles when `ctrl-c` which prevents cleanup.

I added support for this using the `ctrlc` crate.

alexrutar:
@wupr I think I have now resolved all of your concerns with the previous implementation using filesystem locks. The code is quite a bit simpler as well.

1. Now we use a file `.autobib-format`, instead of empty directories
2. We always acquire a shared lock on `.autobib-format` at startup
3. We acquire an exclusive lock when migrating, so that migration can never conflict with adding attachments, `autobib find`, etc.

wupr:
Thanks! I'm glad the lock didn't complicate things. I'll take a closer look later today

alexrutar:
1. I have a bunch of unrelated but substantial upstream changes, which I am not pushing since this PR is already big, but I will have to put changes/fixes there to avoid conflicts
2. I realize even if we fix the format that there is a lot of value in the format / lock file, since it can be used (via `flock(2)`) to coordinate with Autobib when reading the attachments directory. Because of this perhaps using `.autobib-lock` is better? Or some other name, if you have good ideas.

src/app/cli.rs
```diff
@@ -154,6 +154,13 @@ pub enum Command {
         #[arg(short, long)]
         force: bool,
     },
+    /// Backup the record database.
+    ///
+    ///
```
wupr: Some extra blank lines here

src/app/cli.rs
```diff
@@ -275,7 +290,7 @@ pub enum Command {
     /// Import records from BibTeX files.
     ///
     /// The implementation automatically determines a remote identifier from the data using
-    /// your `preferred_providers` config setting, or with any other remote identifier that can
+    /// your `preferred_jeys` config setting, or with any other remote identifier that can
```
wupr:
```suggestion
    /// your `preferred_keys` config setting, or with any other remote identifier that can
```

src/app/cli.rs
```diff
@@ -179,6 +186,9 @@ pub enum Command {
         /// Also delete aliases.
         #[arg(long)]
         delete_aliases: bool,
+        /// Do not warn on orphaned attachment directories.
+        #[arg(short = 'A', long)]
+        ignore_attachments: bool,
```
wupr: This flag is potentially confusing as the flag for `replace` with the same name does a different thing. There the default there is to move attachments, so it could lead the user to think that the default here is to delete attachments. I think something like `--no-attachment-warning` would be better here.

src/app/cli.rs
```diff
@@ -237,6 +247,11 @@ pub enum Command {
         #[arg(short = '1', long, conflicts_with = "limit")]
         one: bool,
     },
+    /// Utilities for maintenance and cleanup.
+    Gc {
```
wupr: `gc` would be a strange name for people who haven't encountered this in programming. Let's call it `clean` or `prune`

src/app.rs
```diff
@@ -319,46 +398,48 @@ pub fn run_cli<C: Client>(cli: Cli, client: &C) -> Result<()> {
...
-                        key,
-                        row: RecordRow { data, .. },
-                    },
-                    row,
+                    record,
+                    // Record {
+                    //     key,
+                    //     row: RecordRow { data, .. },
+                    // },
+                    state,
```
wupr: This can be cleaned up

docs/changelog/next.md
```diff
@@ -1,38 +1,61 @@
...
 
+- `zbmath` identifiers are now stored internally without 0-padding to length 8
 - SQLite is now only bundled when the Cargo feature `bundled-sqlite` is enabled.
   This feature is enabled by default, but this may cause breakage with builds using `--no-default-features`.
   Disabling this feature will cause the compiled binary to link to your SQLite system library instead.
-- Renamed `autobib util list` to `autobib util print-identifiers`.
-  `autobib util list` is still usable as an alias, but this will be removed in the future.
-- Aliases can no longer contain control characters.
+- In a future version, Autobib will migrate to a new attachment folder format.
+  The new folder format is not compatibile with Autobib versions `< 0.7.0`.
```
wupr:
```suggestion
  The new folder format is not compatible with Autobib versions `< 0.7.0`.
```

docs/changelog/next.md
```diff
@@ -1,38 +1,61 @@
...
+  - This version is able to read both the legacy format and the new format.
+  - You can migrate early by running `autobib gc attachments --migrate`.
+    Note that attachments will no longer be readable by old version of Autobib.
+- Aliases can no longer contain control characters (such as tab `\t` or newline `\n`).
```
wupr: Shall we put this, the `util` changes, the `find` mode change, and the `preferred_keys` change into a new section about deprecations and future breaking changes?

docs/changelog/next.md
```diff
@@ -1,38 +1,61 @@
...
 - `autobib find --mode canonical-id` has been renamed to `autobib find --mode records`.
   The CLI still accepts the old name as an alias, but this will be removed in the future.
 - `autobib get` has been reworked.
   Output is no longer sorted and deduplicated: instead, one record is printed for each identifier.
-  This also means that the `--out` and `--append` options have been removed.
-  This behaviour is now included in `autobib source`.
+  This also means that the `--out` and `--append` options have been removed, and warnings are no longer printed for duplicate records.
   There are also new features:
   - Added `--template` option, to format records with an arbitrary template instead of as BibTeX.
   - Added the ability to read from stdin.
```
wupr: I think these should be explained in the new features section instead

docs/changelog/next.md
```diff
@@ -1,38 +1,61 @@
...
+  - `preferred_keys` is more general and can contain a list of regexes to match keys
+  - Migrate `preferred_providers` to `preferred_keys` by replacing each `provider` with the corresponding regex `^provider:.*`
 
 ## New features
 
-- New command `autobib format` which prints records using the template syntax.
 - New template metas:
   - `%json`, a JSON dictionary of all of the available data
   - `%modified`, the modification time of the record
+  - `%key`, the original citation key from the request
```
wupr, line 45: This new meta is not mentioned in the error message for `KeyParseErrorKind::InvalidMeta`

src/path_hash.rs
```diff
@@ -1,43 +1,343 @@
...
+            .truncate(false)
+            .open(root.join(Self::fmt_file()))?;
+        if EXCLUSIVE {
+            fmt.lock()?;
+        } else {
+            fmt.lock_shared()?;
+        };
+
+        Ok(Self { root, fmt })
+    }
```
wupr: The lock cannot be acquired if `root` doesn't exist. This will fail with message:
```
error: No such file or directory (os error 2)
```

src/app/replace.rs
```diff
@@ -96,5 +96,26 @@ where
             .commit()?;
     }
 
-    Ok(())
+    // try to migrate attachments, or warn on orphaned
+    let Some(root) = root else {
+        return Ok(());
+    };
+
+    match root.rename(
```
wupr, line 104: We should also check whether there are attachments to migrate. Otherwise, this would create an empty directory for the merged record if neither of the original records had attachments.

src/app/attach/migrate.rs
```diff
@@ -0,0 +1,199 @@
...
+            )
+        }),
+        5 => true, // base-32 check happens when decoding later on
+        _ => false,
+    }
+}
+
+fn decode_attachment_dir_id(provider: &str, source: &Path) -> Option<RemoteId> {
+    if let Some(encoded) = source.file_name().and_then(|name| name.to_str())
+        && let Ok(sub_id_bytes) = BASE32.decode(encoded.as_bytes())
```
wupr: This will fail to decode v1 attachment paths because their encoding is `BASE32_NOPAD`

src/app/delete.rs
```diff
@@ -57,24 +58,35 @@ pub fn hard_delete<F: FnOnce() -> Vec<(regex::Regex, String)>>(
...
     match record_db.state_from_record_id(id, &config.alias_transform)? {
-        RecordIdState::Entry(original_name, _, state) => entry_callback(original_name, state)?,
-        RecordIdState::Deleted(original_name, _, state) => deleted_callback(original_name, state)?,
-        RecordIdState::Void(original_name, _, state) => voided_callback(original_name, state)?,
+        RecordIdState::Entry(record, state) => {
+            entry_callback(record.key, state)?;
+            return Ok(Some(record.row.canonical));
+        }
+        RecordIdState::Deleted(record, state) => {
+            deleted_callback(record.key, state)?;
```
wupr, line 80: Do we want to return the key here so that `app.rs:345` can warn about orphaned attachments?

Currently, if one hard-deletes a soft-deleted record with attachments, there are no warnings. If we add a `return` here then the warning will also appear when soft-deleting a soft-deleted record (in addition to the error), but I think that's fine, if not desirable.

src/path_hash.rs
```diff
@@ -1,43 +1,343 @@
...
+// A directory `$AUTOBIB_ATTACHMENTS_DIRECTORY/.autobib-format` contains marker directories which
+// describe the current state. The first case below to match determines the behaviour:
+//
+// 1. `.autobib-format` does not exist, or `.autobib-format/v0` exists: this is the legacy
+//    attachment format, using `rapidhash::v1` and `zbmath` identifiers with 0-padding.
+// 2. `.autobib-format/v1-migrating` exists: migration from `v0` to `v1` was interrupted, so the
+//    directories are mixed between the `v0` and `v1` formats.
+// 3. `.autobib-format/v1` exists: this is the new attachment format, using
+//    `rapidhash::v3` and `zbmath` identifiers without 0-padding.
+// 4. Else: the attachment format is unknown to the current binary, resulting in an error
```
wupr: This is out of date now

src/path_hash.rs
```diff
@@ -1,43 +1,343 @@
...
+
+/// In order to reduce the number of files which are in the same directory, we apply a 30-bit
+/// header to each path, which is encoded in base32 as `xx/xx/xx`. Then the corresponding path is:
+/// ```text
+/// provider/xx/xx/xx/base32-encoding-of-sub-id/
+/// ```
+///
+/// The 30 bit header is formed by converting the u64 output of the relevant rapidhash algorithm
+/// applied to the format-specific sub-id into little endian bytes, then taking the four most
+/// significant bytes (in decreasing order), encoding using BASE32 into 8 ASCII characters, and then
```
wupr: I didn't notice this before, but I think the code is taking the 4 _least_ significant bytes?

The number of encoded characters is now 7 without padding

src/app/attach/cleanup.rs
```diff
@@ -0,0 +1,63 @@
...
+    for entry in fs::read_dir(attachment_root.dir())? {
+        let entry = entry?;
+
+        // providers are always ascii alphabetic
+        if entry.file_type()?.is_dir()
+            && entry
+                .file_name()
+                .as_os_str()
+                .to_str()
+                .is_some_and(|s| s.bytes().all(|b| b.is_ascii_alphabetic()))
```
wupr: Should this match `migrate.rs:103-106`?

src/app/attach/migrate.rs
```diff
@@ -0,0 +1,199 @@
...
+            if matches!(
+                err.kind(),
+                io::ErrorKind::AlreadyExists | io::ErrorKind::DirectoryNotEmpty
+            ) =>
+        {
+            error!(
+                "Target directory already exists.\n\
+                 source: {}\n\
+                 target: {}\n\
+                 Merge, remove, or rename one of these directories, then rerun `autobib gc attachemnts --migrate`.",
```
wupr:
```suggestion
                 Merge, remove, or rename one of these directories, then rerun `autobib gc attachments --migrate`.",
```

src/path_hash.rs
```diff
@@ -1,43 +1,343 @@
...
+            lock: lock.upgrade()?,
+            format,
+        })
+    }
+}
+
+impl AttachmentRoot<true> {
+    /// Change the format.
+    pub fn set_format(&mut self, new: AttachmentFormat) -> Result<(), io::Error> {
+        self.lock.set_format(new)
```
wupr: I think we need to set `self.format` as well?

src/config/default_config.toml
```diff
@@ -1,8 +1,18 @@
...
-# - the printed key when selecting an option with `autobib find`
-# - the provider used to attempt remote resolution of imported bibtex entries
+# This value is deprecated and has been replaced by `preferred_keys`. To convert
+# the value, replace each `provider` with the regex `^provider:.*`. For example,
+#   preferred_providers = ["zbl", "mr"]
+# is equivalent to
+#   preferred_keys = ["zbl:", "mr:"]
 preferred_providers = []
 
+# Regexes for automatic selection of keys. This is used to choose:
```
wupr, line 8: We should link to a page that documents the regex syntax used ~~(since there are so many variants). Maybe https://docs.rs/regex?~~

Update: we did this for `alias_transform`, but it's probably worth repeating here

src/app/delete.rs
wupr: The suggestion is out of date

wupr (changes requested):

wupr:
> 2\. Because of this perhaps using `.autobib-lock` is better? Or some other name, if you have good ideas.

I'd be happy with `.autobib-lock`

wupr:
Thanks for implementing these big changes. I tried to keep an eye out for inconsistencies when reviewing, since I know they are hard to eliminate when you're working on big changes.

I guess the one thing I haven't yet tried is the new template keys. Update: Tried them now; amazing

src/app/cli.rs
alexrutar: I like `clean`.

alexrutar:
> Thanks for implementing these big changes. I tried to keep an eye out for inconsistencies when reviewing, since I know they are hard to eliminate when you're working on big changes.
> 
> I guess the one thing I haven't yet tried is the new template keys. Update: Tried them now; amazing

No problem, thanks for the careful review! I will continue to go over the PR.

alexrutar:
(And merge soon, when I am also happy with it.)

docs/changelog/next.md
alexrutar: Good idea!

src/path_hash.rs
alexrutar: I didn't see anywhere that `root` is not created? Did you spot somewhere?

src/app/attach/migrate.rs
alexrutar: Yikes, I changed `BASE32` to `BASE32_NOPAD` but forgot to change it here.

src/path_hash.rs
alexrutar: Yes, least significant, I get this backwards approximately 50% of the time.

src/app/replace.rs
alexrutar: While thinking about how this interacts with atomicity I noticed that there is an interaction between the filesystem locks and sqlite transactions: I think the filesystem lock needs to be longer than the sqlite transaction, and for `replace` we actually need an exclusive lock. I think we need to:

1. Acquire attachment lock
2. Rename the record in the db
3. Rename of required
4. Release attachment lock

~~If not, another thread could witness that the attachment is renamed but the record has not been replaced yet.~~

~~The reasoning for the current order is that if there is an issue with the attachments (or any issue in general, even if it is not a DB error) the transaction will roll back. With the above order the attachment could be renamed, but then there is some other issue, and the transaction fails to commit, and then we are in an even worse state.~~

Edit: This is wrong, it could happen with either order. Will think if there is any way to do this properly, but otherwise I think we just leave it as 'best-effort'.

src/path_hash.rs
alexrutar: Nevermind I can see where

alexrutar:
Almost everything is fixed, just need to think about filesystem locks more. Outstanding things to fix:

1. The `cli.read_only` flag is not respected if `.autobib-lock` does not exist, since it is created no matter what. If we use `cli.read_only` we should print a warning that `.autobib-lock` does not exist, so a shared lock could not be acquired, and this means that some reads might be corrupted.
2. Think carefully if there are places where exclusive filesystem lock would be better than shared lock. (Maybe `autobib import`)
3. Think if there is an easy way for autobib to share exclusive locks with other processes. A common pattern might be: parent process acquires an exclusive lock, and then does a lot of work including some operations using `autobib`. E.g., we could add the option to open the attachment directory without using any locks.

Note 1: I pushed a WIP commit for 1, but maybe a good solution for 2 and 3 will require different implementation.

Note 2: Eventually, we will be able to assume existence of `.autobib-lock` as a requirement to even read the attachments directory, so we don't have to come up with a robust solution now.

src/app/replace.rs
alexrutar: Anyway I implemented your suggestion as a 'best-effort' fix.

alexrutar:
@wupr Would be happy for your input on the lock situation too, but anyway r.e. above:

1. I implemented best-effort checks
2. I added `autobib import --include-files`, even if it isn't strictly necessary. I did this since in #314 an import could take a long time, and it would be inconvenient if the attachment directory changed a lot during this process.  I did not add it for `autobib attach`, since it is a single file which is "close to atomic", whatever that means. We should eventually decide what we actually mean by "shared" and "exclusive", but we don't have to do that now since we aren't even committing to locks in general yet.
3. Will hold off on this for now; I created #327

src/app/cli.rs
```diff
@@ -641,7 +646,7 @@ impl Command {
             Self::Update { .. } => "update",
             Self::Edit { .. } => "edit",
             Self::Hist { .. } => "hist",
-            Self::Gc { .. } => "gc",
+            Self::Clean { .. } => "gc",
```
wupr: "gc" -> "clean"

tests/cli.rs
```diff
@@ -10,6 +10,8 @@ use std::{
     path::{Path, PathBuf},
 };
 
+static AUTOBIB_LOCKFILE: &str = ".autobib-lock";
```
wupr: I thought a bit more about this – maybe `.autobib` would be better? As it serves as more than just a lock.

src/app/cli.rs
```diff
@@ -413,14 +422,10 @@ pub enum Command {
     /// the database, its data will be retrieved first.
     Replace {
         /// The identifier to replace.
-        #[clap(value_name = "IDENTIFIER")]
+        #[clap(value_name = "KEY")]
```
wupr, line 425: This made me realise something – are we relaxing the rules on what is called a key?

(I know the renaming here is probably to match the `%key` meta and `preferred_keys` config.)

docs/changelog/next.md
```diff
@@ -1,38 +1,63 @@
...
+
+## Deprecations and future breaking changes
+
+- In a future version, Autobib will migrate to a new attachment folder format.
+  The new folder format is not compatible with Autobib versions `< 0.7.0`.
+  - This version is able to read both the legacy format and the new format.
+  - You can migrate early by running `autobib clean attachments --migrate`.
+    Note that attachments will no longer be readable by old version of Autobib.
+- Aliases can no longer contain control characters (such as tab `\t` or newline `\n`).
+  - Existing aliases containing control characters can still be accessed and renamed.
```
wupr: I think this is an immediate breaking change, while the `autobib find` and `autobib util list` changes are future breaking changes.

wupr (commented):
I don't have much to say about the lock situation; I think your implementation is very sensible. Exclusive lock for long operations (migration, clean-up, import with files), and shared locked for things that shouldn't happen concurrently to a long operation.

wupr (approved):
Hmm, is my previous "Request changes" blocking the merge? You can probably bypass the block, but I'll make it easier anyway.

tests/cli.rs
wupr: Actually, I'm not sure any more, since `.autobib` is also proposed as a directory for local config files in #241

alexrutar: ~~Maybe `.autobib/LOCK`?~~

alexrutar: I got totally confused, of course it's fine since this is inside the attachment directory. I think keeping it as .autobib-lock is fine since eventually it will only be a lock file.

wupr: > Maybe `.autobib/LOCK`?

I think this is a good option. Would it complicate things though if the lock file is one directory deeper?

src/app/cli.rs
alexrutar: I've got some ideas for renaming (also related to that draft PR about a glossary from a while back), let's revisit this in another PR. Also a good chance to use an LLM to try to come up with good terminology, and check the whole codebase for consistency.

tests/cli.rs
wupr: > I got totally confused, of course it's fine since this is inside the attachment directory. I think keeping it as .autobib-lock is fine since eventually it will only be a lock file.

Sure. I just thought consistency is nice. But given that this is the only place where `.autobib-lock` would exist, I guess it's OK.

Also, GitHub hasn't been pulling new comments to the page. Good thing I refreshed

alexrutar:
> Hmm, is my previous "Request changes" blocking the merge? You can probably bypass the block, but I'll make it easier anyway.

It's not very hard to bypass, but yes it's a tiny bit easier if approved! Thanks

PR #284, opened 2026-05-06
Release version 0.6.1

alexrutar:
I thought it would be good to release a patch version which includes the zbmath fixes, independently of changing the internal format which we should do in `v0.7.0` (if at all).

@wupr

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -114,7 +114,7 @@ dependencies = [
 
 [[package]]
 name = "autobib"
-version = "0.6.0"
+version = "0.6.1"
 dependencies = [
  "anyhow",
  "assert_cmd",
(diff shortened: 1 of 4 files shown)


wupr (approved):
Good idea! I couldn't leave comments on the diff (GitHub says "internal error"), so I'll put them here.

docs/changelog/v0.6.1.md‎, L3: `v.0.5.1` --> `v0.6.0`
docs/changelog/v0.6.1.md‎, L9: (nitpick) this is the only item in the list that is not a verbal phrase

alexrutar:
Thanks! Yeah GitHub has been acting up a lot recently. I heard that some open source projects are moving off of it because of this.

PR #286, opened 2026-05-08
Fix SHA-256 command in shell scripts

wupr:


wupr:
I've been using my own script on my Mac, but when I tried the committed testing script today on a new machine (Linux), I encountered the error that `sha256` is not found. I don't think it's a standard command on Linux ~~or macOS~~. Is it a shell alias that you have?

alexrutar:
Interesting! Yes, `sha256` is pre-installed on my mac at `/sbin/sha256`. I agree that `shasum -a 256` is better.

alexrutar (approved):

alexrutar:
> I've been using my own script on my Mac

Out of curiosity do you find that the existing script is lacking in some way?

wupr:
> Out of curiosity do you find that the existing script is lacking in some way?

No, quite the contrary. I rely on editor extensions (and I guess GitHub Actions) for most of the checking. My script does very little

PR #287, opened 2026-05-10
Allow compiling without bundled SQLite

### LINKED ISSUE #285: test failure "delete"
Tested in 0.6.1:

```
failures:

---- delete stdout ----

thread 'delete' (236) panicked at /rustc/59807616e1fa2540724bfbac14d7976d7e4a3860/library/core/src/ops/function.rs:250:5:
Unexpected stderr, failed var.contains(Cannot delete undefined alias)
├── var: error: Query returned no rows
└── var as str: error: Query returned no rows

command=`"/builddir/autobib-0.6.1/target/x86_64-unknown-linux-gnu/release/autobib" "--database" "/tmp/.tmpTTs2r5/records.db" "--config" "/tmp/.tmpAm1kiA/config.toml" "--attachments-dir" "/tmp/.tmp0P1TVY" "--no-interactive" "delete" "--hard" "local:first" "my_alias"`
code=1
stdout=""
stderr="error: Query returned no rows\n"

note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    delete

test result: FAILED. 49 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 9.86s
```

alexrutar:
This makes it possible to compile without bundling SQLite.

Bundling SQLite is now behind a feature flag called `bundled-sqlite` which is enabled by default.

If SQLite is not bundled, some additional things are done at runtime:

1. enabling foreign keys (which is not done by default since the bundled SQLite copy is compiled with a flag to always enable foreign keys on startup)
2. checking that the runtime SQLite is >= 3.35

I also updated the test script to check for `AUTOBIB_USE_SYSTEM_SQLITE = 1` which then builds the response cache / runs the tests with system SQLite.

Fixes #285 (and provides a more reasonable option downstream)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,7 @@ Make sure the [rust toolchain](https://www.rust-lang.org/tools/install) is insta
 ```sh
 cargo install --locked autobib
 ```
+See the [build docs](docs/build.md) for more options.
 
 ## Basic usage
 
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,27 @@
+fn main() {
+    println!("cargo::rerun-if-env-changed=LIBSQLITE3_SYS_USE_PKG_CONFIG");
+
+    let bundled_sqlite = std::env::var_os("CARGO_FEATURE_BUNDLED_SQLITE").is_some();
+    let use_pkg_config =
+        std::env::var_os("LIBSQLITE3_SYS_USE_PKG_CONFIG").map(|value| value != "0");
+
+    match (bundled_sqlite, use_pkg_config) {
+        (true, Some(true)) => {
+            println!(
+                "cargo::error=`LIBSQLITE3_SYS_USE_PKG_CONFIG` requests system SQLite via \
+                 pkg-config, but the `bundled-sqlite` Cargo feature is enabled. Set \
+                 `LIBSQLITE3_SYS_USE_PKG_CONFIG=0` or disable the `bundled-sqlite` feature."
+            );
+            std::process::exit(1);
+        }
+        (false, Some(false)) => {
+            println!(
+                "cargo::error=`LIBSQLITE3_SYS_USE_PKG_CONFIG=0` disables system SQLite via \
+                 pkg-config, but the `bundled-sqlite` Cargo feature is disabled. Set \
+                 `LIBSQLITE3_SYS_USE_PKG_CONFIG=1` or enable the `bundled-sqlite` feature."
+            );
+            std::process::exit(1);
+        }
+        _ => {}
+    }
+}
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -19,21 +19,27 @@ CACHE_DIR="${CACHE_ROOT}/test-cache-${HASH}"
 
 export AUTOBIB_RESPONSE_CACHE_PATH="${CACHE_DIR}/responses.dat"
 
+if [[ "${LIBSQLITE3_SYS_USE_PKG_CONFIG:-0}" != "0" ]]; then
+    FEATURE_ARGS=(--no-default-features)
+else
+    FEATURE_ARGS=()
+fi
+
 if [[ ! -f "${AUTOBIB_RESPONSE_CACHE_PATH}" ]]; then
     echo 2>&1 "Cache file not found. Generating cache file: ${AUTOBIB_RESPONSE_CACHE_PATH}"
 
     mkdir -p "${CACHE_DIR}"
 
     # Generate the cache file
-    cargo run --locked --features write_response_cache -- -vv source --retrieve-only --ignore-null "${REMOTES_FILE}"
+    cargo run --locked "${FEATURE_ARGS[@]}" --features write_response_cache -- -vv source --retrieve-only --ignore-null "${REMOTES_FILE}"
 else
     echo 2>&1 "Cache file found: ${AUTOBIB_RESPONSE_CACHE_PATH}"
 fi
 
-cargo test --locked --no-fail-fast --features read_response_cache -- "$@"
+cargo test --locked --no-fail-fast "${FEATURE_ARGS[@]}" --features read_response_cache -- "$@"
 
-cargo doc --no-deps --locked
-cargo clippy --locked
+cargo doc --no-deps --locked "${FEATURE_ARGS[@]}"
+cargo clippy --locked "${FEATURE_ARGS[@]}"
 cargo fmt --check
 sort -C "${REMOTES_FILE}"
 REPETITIONS="$(uniq -d < "${REMOTES_FILE}" | wc -w)"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ rust-version = "1.94"
 edition = "2024"
 
 [features]
+default = ["bundled-sqlite"]
+bundled-sqlite = ["rusqlite/bundled"]
 in_memory_database = []
 read_response_cache = ["dep:bincode"]
 write_response_cache = ["dep:bincode", "in_memory_database"]
@@ -40,7 +42,7 @@ ramify = "0.9.0"
 regex = "1.11"
 regex-syntax = "0.8"
 rsxiv = { version = "0.4.3", features = ["serde"] }
-rusqlite = { version = "0.39", features = ["bundled", "chrono", "functions"] }
+rusqlite = { version = "0.39", features = ["chrono", "functions"] }
 serde = { version = "1.0", features = ["derive"] }
 serde_bibtex = "0.7.1"
 serde_json = "1.0"
--- a/docs/build.md
+++ b/docs/build.md
@@ -1,0 +1,40 @@
+# Build configuration
+
+## Bundling SQLite
+
+By default, Autobib compiles with a bundled copy of SQLite enabled by the `bundled-sqlite` Cargo feature.
+To link against the SQLite library available on your system instead, disable this feature:
+
+```sh
+cargo install --locked autobib --no-default-features
+```
+This makes the binary about 1.5MB smaller, at the cost of potential compatibility issues.
+Note that the system SQLite library must be version 3.35.0 or newer.
+If this is not the case, Autobib will fail with a runtime error.
+
+Note that Autobib is only tested against the bundled SQLite configuration by default, so using the system SQLite copy may result in difficult-to-diagnose errors.
+You can run the test script to check compatibility with your system SQLite library using
+```sh
+LIBSQLITE3_SYS_USE_PKG_CONFIG=1 ./scripts/test.sh
+```
+Note that the bundled copy of SQLite is compiled with the following flags:
+```sh
+SQLITE_DEFAULT_MEMSTATUS=0
+SQLITE_DEFAULT_WAL_SYNCHRONOUS=1
+SQLITE_DQS=0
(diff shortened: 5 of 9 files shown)


wupr:
Does this mean we want to add [sqlite](https://formulae.brew.sh/formula/sqlite) as a dependency in our Homebrew formula?

alexrutar:
> Does this mean we want to add [sqlite](https://formulae.brew.sh/formula/sqlite) as a dependency in our Homebrew formula?

No, it doesn't change the default build (e.g. what you get when you run "cargo install"), or what's in the release binaries. It just makes it possible to build without SQLite if you really want.

wupr:
> No, it doesn't change the default build (e.g. what you get when you run "cargo install"), or what's in the release binaries. It just makes it possible to build without SQLite if you really want.

Mhmm, but do we want to use the unbundled build? If you have Python or Deno installed with Homebrew, then you already have the sqlite formula installed. On the other hand, I don't know how easy it is to do dynamic linking in Homebrew across macOS and Linux

alexrutar:
Ah I understand what you mean! I still think bundling is better since it means there are definitely no SQLite compatibility issues; on top of using a recent version we're also using lots of compilation flags.

tornaria:
Works for me. I wonder if the `pragma_update()` call should be done unconditionally, to avoid any trouble.

As things are now, if I build with `LIBSQLITE3_SYS_USE_PKG_CONFIG=1 cargo build --release --locked` I end up with a binary that uses system sqlite but does not enable the foreign key option.

Also, maybe there are other options in `LIBSQLITE3_FLAGS` that can be configured at run-time without any cost (either instead, or in addition). Since our sqlite package is for general use, there are some options and defaults we cannot change.

alexrutar:
> Works for me. I wonder if the `pragma_update()` call should be done unconditionally, to avoid any trouble.
> 
> As things are now, if I build with `LIBSQLITE3_SYS_USE_PKG_CONFIG=1 cargo build --release --locked` I end up with a binary that uses system sqlite but does not enable the foreign key option.

I thought about this and I understand the point, but felt that for 'unofficial' build patterns we should not go out of our way to try to guess what needs to be done in order to make all strategies work (in the `bundled` configuration). Adding environment variables which impact compilation can basically always cause breakage in some way. One possibility I would be more open to is to check (at compile time) that such flags are not set and to raise errors, e.g. so that running `LIBSQLITE3_SYS_USE_PKG_CONFIG=1 cargo build --release --locked` results in an error.

@wupr do you have any opinions about this?

> Also, maybe there are other options in `LIBSQLITE3_FLAGS` that can be configured at run-time without any cost (either instead, or in addition). Since our sqlite package is for general use, there are some options and defaults we cannot change.

Outside the bundled configuration I think it is totally reasonable to be more defensive! I will look at it; also if you spot anything I would be happy to hear it.

tornaria:
> I thought about this and I understand the point, but felt that for 'unofficial' build patterns we should not go out of our way to try to guess what needs to be done in order to make all strategies work (in the `bundled` configuration). Adding environment variables which impact compilation can basically always cause breakage in some way.

I understand. As a counter point, note this is not my choice of build pattern. This is distro policy, every rust package gets built with `LIBSQLITE3_SYS_USE_PKG_CONFIG=1` and other similar variables to avoid vendoring of dependencies as much as possible.

In this particular case, what would be the downside of enabling `foreign_keys` at runtime?
If you need foreing_keys option enabled, what would be the downside of enabling it at startup?

> One possibility I would be more open to is to check (at compile time) that such flags are not set and to raise errors, e.g. so that running `LIBSQLITE3_SYS_USE_PKG_CONFIG=1 cargo build --release --locked` results in an error.

Sure: testing did its job, we caught the issue, and I won't ship a broken binary. But if you can catch broken combination of flags and warn/error at compile time, so much better. OTOH, if you can catch `LIBSQLITE3_SYS_USE_PKG_CONFIG=1` maybe you can just disable the `bundled-sqlite` feature?

scripts/test.sh
```diff
@@ -19,21 +19,27 @@ CACHE_DIR="${CACHE_ROOT}/test-cache-${HASH}"
 
 export AUTOBIB_RESPONSE_CACHE_PATH="${CACHE_DIR}/responses.dat"
 
+if [[ "${AUTOBIB_USE_SYSTEM_SQLITE:-0}" == "1" ]]; then
+    SQLITE_ARGS=(--no-default-features)
```
wupr: I would call this variable `FEATURE_ARGS` since it is potentially related to more than just sqlite (when we have more default features). But it doesn't really matter right now.

wupr (approved):

wupr:
> do you have any opinions about this?

@alexrutar I think I agree with you. If we always enable "foreign_keys" on startup, then one could make the argument that we should also always check the sqlite version, and we'd find ourselves on the path of trying to accommodate all "unofficial" build options.

wupr:
> OTOH, if you can catch `LIBSQLITE3_SYS_USE_PKG_CONFIG=1` maybe you can just disable the `bundled-sqlite` feature?

I'm not sure if you can disable a feature in build.rs, but one could potentially detect the environment variable and set a cfg flag that is used in `db.rs` instead of `not(feature = "bundled-sqlite")`.

alexrutar:
I agree with @wupr that seems a bit weird to dynamically change feature flags in `build.rs`. I think having a separate build variable entirely is also a bit weird. In the end what I've implemented in `build.rs` is to check if the `LIBSQLITE3_SYS_USE_PKG_CONFIG` value is compatible with the cargo feature, and to fail the build with an error if it is not.

scripts/test.sh
alexrutar: You mean, for the variable `FEATURE_ARGS` to (essentially) include a list of feature flags? I guess I wanted a slightly different interface than 'pass the arguments' since we also need to manipulate the feature args directly inside the script (with the `read-response-cache` and `write-response-cache` features) and I thought the interface should reflect this.

Maybe the best choice here is actually to just read `LIBSQLITE3_SYS_USE_PKG_CONFIG` instead?

alexrutar: Oh you meant the `SQLITE_ARGS` variable of course! That makes sense to me.

alexrutar: For consistency with `build.rs` I also changed `AUTOBIB_USE_SYSTEM_SQLITE` to `LIBSQLITE3_SYS_USE_PKG_CONFIG` so this also means when running the script on a system with `LIBSQLITE3_SYS_USE_PKG_CONFIG` set by default the test will run correctly.

wupr: > Oh you meant the `SQLITE_ARGS` variable of course! That makes sense to me.

Yes, I meant to highlight just line 23, sorry.

README.md
```diff
@@ -343,6 +343,36 @@ Autobib supports basic global configuration through a [TOML](https://toml.io/) f
...
+Note that the system SQLite library must be version 3.35.0 or newer.
+If this is not the case, Autobib will fail with a runtime error.
+
+Note that Autobib is only tested against the bundled SQLite configuration by default, so using the system SQLite copy may result in difficult-to-diagnose errors.
+You can run the test script with your system SQLite library using
+```sh
+AUTOBIB_USE_SYSTEM_SQLITE=1 ./scripts/test.sh
+```
+The bundled copy of SQLite is compiled with the following flags:
+```sh
```
wupr: I think `txt` would be better

README.md
```diff
@@ -343,6 +343,36 @@ Autobib supports basic global configuration through a [TOML](https://toml.io/) f
...
+cargo install --locked autobib --no-default-features
+```
+This makes the binary about 1.5MB smaller, at the cost of potential compatibility issues.
+Note that the system SQLite library must be version 3.35.0 or newer.
+If this is not the case, Autobib will fail with a runtime error.
+
+Note that Autobib is only tested against the bundled SQLite configuration by default, so using the system SQLite copy may result in difficult-to-diagnose errors.
+You can run the test script with your system SQLite library using
+```sh
+AUTOBIB_USE_SYSTEM_SQLITE=1 ./scripts/test.sh
```
wupr: This has been changed now. Also needs to be updated in `tests/README.md`

README.md
```diff
@@ -343,6 +343,36 @@ Autobib supports basic global configuration through a [TOML](https://toml.io/) f
 This path can be modified with the `AUTOBIB_CONFIG_PATH` environment variable.
 You can generate a default configuration file with `autobib default-config`, or view the configuration options [here](src/config/default_config.toml).
 
+### Bundling SQLite
```
wupr: I think we should keep custom build options like this in `docs/`, with a link from here.

build.rs
```diff
@@ -0,0 +1,27 @@
...
+                 pkg-config, but the `bundled-sqlite` Cargo feature is enabled. Set \
+                 `LIBSQLITE3_SYS_USE_PKG_CONFIG=0` or disable the `bundled-sqlite` feature."
+            );
+            std::process::exit(1);
+        }
+        (false, Some(false)) => {
+            println!(
+                "cargo::error=`LIBSQLITE3_SYS_USE_PKG_CONFIG=0` disables system SQLite via \
+                 pkg-config, but the `bundled-sqlite` Cargo feature is disabled. Set \
+                 `LIBSQLITE3_SYS_USE_PKG_CONFIG=0` or enable the `bundled-sqlite` feature."
```
wupr:
```suggestion
                 `LIBSQLITE3_SYS_USE_PKG_CONFIG=1` or enable the `bundled-sqlite` feature."
```

wupr (commented):
Not sure if you're still working on this, but since I took another look I'll just leave some comments

README.md
alexrutar: I thought `sh` was good, because it makes the variable highlighting look slightly nicer?

alexrutar:
Thanks for the careful review! I've fixed everything you mentiond; I moved the docs to `docs/build.md` and also added the details on dynamic linking for musl.

alexrutar:
This should go into `v0.7.0` since technically it is breaking.

alexrutar:
I'm going to merge this for now and work on #282 but if there are any more problems we can discuss more in a new PR / issue page.

README.md
wupr: I think `SQLITE_LIKE_DOESNT_MATCH_BLOBS` etc get rendered like commands because they don't have `=`. On the website code blocks might have their language shown explicitly, so `sh` would be wrong (we can suppress the language of `txt` blocks).

alexrutar: Isn't the language also (sort of) correct? Setting variables like this is very much a shell-ism. I could imagine a (rather boring) shell script that just defines a bunch of variables.

At the same time I also agree that txt is quite correct!

wupr: What is bothering me is that it is not a correct shell script, because `SQLITE_LIKE_DOESNT_MATCH_BLOBS` without `=` is not valid syntax for setting a variable (it would result in a command-not-found error).

I'm also wondering if this should actually be a markdown list

wupr:
> This should go into `v0.7.0` since technically it is breaking.

Well, I imagine this is going to released in Void Linux as a patched version of v0.6.1

README.md
alexrutar: Okay, yes I agree markdown list is the best!

alexrutar:
Fair enough, yes! Though for our releases it's fine separately.

PR #288, opened 2026-05-13
Fix build docs markdown formatting

alexrutar:


wupr:
Thanks! Now I can sleep well at night!

PR #289, opened 2026-05-15
Adapt to the custom Homebrew tap's new formula bump process

wupr:
See also autobib/homebrew-autobib#17


alexrutar:
Perhaps now we can publish the release directly (i.e. omit step 9) from the action? If I recall previously the main reason it wasn't automatic was since it triggered the brew workflow.

wupr:
> Perhaps now we can publish the release directly (i.e. omit step 9) from the action? If I recall previously the main reason it wasn't automatic was since it triggered the brew workflow.

Step 9 also gave us the opportunity to retract. We could automate the publication inside a deployment environment?

alexrutar:
At this point, the release is already published to Cargo, so it doesn't seem like too much of a gain in my opinion

wupr:
> At this point, the release is already published to Cargo, so it doesn't seem like too much of a gain in my opinion

Oh you meant for it to come after the crates.io stage? Then sure.

(OK I suppose that is the order in the manual right now. I was thinking of publishing the release just after the artifacts have been uploaded.)

wupr:
That makes a lot of sense actually. I've implemented the suggestion now

alexrutar (approved):
Looks good to me!

Issue #52, opened 2024-09-11
Support Typst


wupr:
For a description of how to do citation using a `.bib` file in Typst, see https://typst.app/docs/reference/model/cite/.

Since Typst is already written in Rust, it is very easy to parse a source file using the [typst-syntax](https://crates.io/crates/typst-syntax) crate, but I don't know if this is time-efficient or memory efficient (compared to the approach in `src/cite_search/tex.rs`).

Either way, a problem is that references to the bibliography and those to labels in the document are not necessarily syntactically distinguished, _i.e._ they can both be of the form `@target` or `#ref(<target>)`; citations can additionally be `#cite(<target>)`.
An obvious strategy is to collect the references that do not exist as labels, plus the explicit `#cite`s.

alexrutar:
Implementation notes:

1. There are multiple ways to cite: either `@ref`, or `#cite<ref>`. In the second case there is a special 'label' syntax `#cite(#label(ref))` but I guess in `typst-syntax` these will reduce to the same element in the syntax tree
2. Since `@ref` could refer to any label, we need to collect all labels in the entire document and only produce citation keys for `@ref`s which do not refer to a label inside the document. Are there other places where there could be labels?
3. How to handle multiple files? In the `.tex` case this is easy since `\cite{...}` is only additive, so we can compute the citation keys for each file separately, and then take the union. I guess in the `.typ` case we need to build three dictionaries: one with all of the `#label`s, one with all of the `@ref`s, and one with all of the `#cite<ref>`s, and then diff right at the very end. What are the cross-file import rules? Should we be providing `import` / `include` auto-discovery or should we mandate that the user pass all relevant files?

wupr:
We could have a single dictionary where the key is the label, and the value tells you whether the label is defined in the document, a citation, or ambiguous. At the end we can filter the keys by the paired value.

My opinion is not made up about auto-discovery of imports, but I think it's fine to require all relevant files be passed, since that matches the .tex behaviour?

alexrutar:
My instinct is to agree with you but I have this concern about the nature of error messages that I can't quite kick; in the `.tex` case if you forget to include one of your `.tex` files, you will just get an 'undefined citation key' error when you try to compile, and it will tell you which `.tex` file this error was in; but in the `.typ` case if you forget to include a file you can get an error from `autobib` like 'undefined alias' or 'invalid remote id' which does not explain the source of the problem.

[...]

PR #292, opened 2026-06-13
Minimal implementation of Typst source

alexrutar:
Implements a basic version of #52 

This uses the `typst-syntax` crate to find all `#cite(...)` and `#cite(label(...))` occurrences. This does not handle `@ref` since I think this would (essentially) require compiling the document in order to handle properly. The problem is that this also does not catch things like
```
#let c = cite
#c(<ref>)
```

I tried to check if `typst query` would be able to extract citekeys programmatically. In principle something like `typst query file.typ cite --field key` should work, but this fails to resolve if the citation key does not exist in the provided bibliography, which is unfortunate.

I looked into plugins a bit, but it seems like Typst prefers plugins to be pure (essentially, text-transformation) which makes network access a bit of a difficulty. We could easily create a WASM plugin which *only* looks in the local database (which can be opened, albeit somewhat inefficiently, using `read`). Should look into this more.


alexrutar:
I asked on the Typst forum as well, and I got a good reply about using `typst eval`: https://forum.typst.app/t/how-to-extract-all-bibliographic-citation-keys-from-a-typst-file/9083/3

PR #298, opened 2026-06-19
Preserve release target rustflags

### LINKED ISSUE #294: RUSTFLAGS override in release.yml
`.cargo/config.toml` sets the `rustflags` for Windows MSVC and Linux musl builds. However they're overridden by a environment variable in the release workflow.

https://github.com/autobib/autobib/blob/0bf82a5a6db97b55be8b164b64e9c8c52989031b/.github/workflows/release.yml#L14

In practice, I think this has only affected the MSVC builds, as the flags in `.cargo/config.toml` agree with Rust's current defaults for the musl targets.

alexrutar:
Possible fix for #294 

I changed this PR: now it sets `-D warnings` in `.cargo/config.toml` globally. It might be slightly annoying for warnings to upgrade to errors whenever one runs `cargo build`, but I think the benefit of all of the configuration being in one place outweighs the downsides.

Previously, this  manually set the `cfg(all)` target when invoking `cargo build --release`. If you have a convincing reason why this is better, I am happy to hear it!

See https://doc.rust-lang.org/cargo/reference/config.html#targetcfgrustflags for documentation of this behaviour.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,6 @@ permissions:
 
 env:
   CARGO_TERM_COLOR: always
-  RUSTFLAGS: -D warnings
   RUSTDOCFLAGS: -D warnings
   RUST_BACKTRACE: 1
   RUST_VERSION: ${{ vars.RUST_VERSION }}
@@ -163,7 +162,9 @@ jobs:
       - name: Build release binary
         shell: bash
         run: |
-          cargo build --verbose --release --locked --target ${{ matrix.target }}
+          cargo build --verbose --release --locked \
+            --config 'target."cfg(all())".rustflags = ["-D", "warnings"]' \
+            --target ${{ matrix.target }}
           if [[ "${{ matrix.image }}" = "windows-latest" ]]; then
             bin="target/${{ matrix.target }}/release/autobib.exe"
           else


alexrutar:
~~Another good option is to just set this in `.cargo/config.toml` so it always applies.~~

This is now what is implemented.

wupr:
I had been thinking we could inject the extra lines in `.cargo/config.toml` in CI using `sed`. But the idea from this StackOverflow answer seemed more elegant: https://stackoverflow.com/a/56873092. Basically build.rs can check if it is in a CI environment (we can use an environment variable to detect this), and toggle a `cfg` which, through `.cargo/config.toml`, adds the `-D warnings` flag (this last part deviates from the cited answer).

This avoids the annoyances we'd have during local development, and I think it's not too ugly. Happy to hear your opinions.

The inability to append to `RUSTFLAGS` seems to have been raised as an issue for a long time without proper resolution: https://github.com/rust-lang/cargo/issues/5376.

alexrutar:
I would hesitate to add an extra dependency just for this, when there are manual (albeit slightly hacky) solutions. Though I understand not wanting to set `-D warnings` globally for all builds, in which case I think it is better to use
```
--config 'target."cfg(all())".rustflags = ["-D", "warnings"]'
```
as was previously implemented.

We can put this into a `$CARGO_EXTRA_RUSTFLAGS` variable declared at the top, the main thing is remembering to include this in every build step. The advantage of putting it directly in `.cargo/config.toml` is that it is always picked up.

If there are multiple cargo config files, they get unified: https://doc.rust-lang.org/cargo/reference/config.html#hierarchical-structure

Maybe there is some more clever solution using there. (We would need slightly different code on the Windows vs Unix runners though).

alexrutar:
Slightly cleaner option: `--config` can accept an extra config file path, instead of manually passing the values at the command line. So, we could create a file `.github/workflows/cargo_extra_config.toml` and put stuff there.

wupr:
> I would hesitate to add an extra dependency just for this,

Indeed. To clarify, if we use an environment variable for CI detection though, then we don't need the extra dependency

alexrutar:
> > I would hesitate to add an extra dependency just for this,
> 
> Indeed. To clarify, if we use an environment variable for CI detection though, then we don't need the extra dependency

Ah, I see what you mean! Set an environment variable in CI only and use `build.rs` for manipulation. Perhaps however it is still simplest (for now) to just set the `--config` directly, and if it gets complicated set extra config in a separate file.

wupr:
> Perhaps however it is still simplest (for now) to just set the `--config` directly, and if it gets complicated set extra config in a separate file.

Yeah, I think I agree. I'd be most happy with this implementation.

> We can put this into a `$CARGO_EXTRA_RUSTFLAGS` variable declared at the top, the main thing is remembering to include this in every build step. The advantage of putting it directly in `.cargo/config.toml` is that it is always picked up.

If we do this in the future, we could also set an alias for `cargo`  in CI that wraps the extra flags inside. (Bash aliases could be slightly more complicated on GitHub than on GitLab...)

> If you have a convincing reason why this is better, I am happy to hear it!

I think it's useful to distinguish between warnings and errors. You can leave some redundant or untidy code (which result in warnings) in until you've decided which code structure works best.

wupr (approved):

PR #308, opened 2026-06-25
Major changes to `autobib get` and `autobib list`.

### LINKED ISSUE #305: Output formatters
We could add an output formatter to `autobib find`, instead of just printing the canonical ID. Maybe there are other places where output formatting would make sense.

Note that, at the cost of repeating lookups, one can use pass the resulting IDs to `autobib format` to emulate this behaviour currently.

alexrutar:
Changes:

- `autobib get` is streaming: records are written to standard output as they are received (unless the record is skipped because of an invalid bibtex entry key, or with the `--strict` flag). This also adds support for a `--template` option and removes `--append` and `--out`.
- I refactored the `%json` meta to now print all of the data in the record row. I also added a `%modified` meta, which prints the modification timestamp.
- I added back `autobib list` but instead with the purpose of listing identifiers in the database (and perhaps formating them). This replaces `autobib util list`, and also adds a `--template` option and an optional position glob argument to filter the identifiers listed.
- I added a `--json` option to `autobib source`. It outputs a dictionary where the keys are the identifiers and the values are the record data. Unlike `%json` in templates, the output is designed to exactly mimic the bibtex data (i.e., does not contain canonical IDs, modification time, etc.).

Internally, I updated the `Record` struct to be a wrapper around a `RecordRow` which also contains a citation key. In particular, a `Record` now also carries the modification time with it.


alexrutar:
@wupr I don't think detailed code review here is needed (there are a lot of code changes, but nothing very complex or critical). But let me know what you think about the API changes!

alexrutar:
Spurious CI failure as usual

wupr:
I think I said this elsewhere (sorry if I didn't), but the API changes all look great!

Issue #328, opened 2026-07-17
Template JSON pointer support


alexrutar:
To reduce dependence on external tools (like `jq`) when working with the `%json` meta, we could support [JSON pointer](https://datatracker.ietf.org/doc/html/rfc6901) syntax, so for instance `{%json/data/fields/year}` is equivalent to `{year}`.

alexrutar:
Alternative syntax: use `.` composition, with JSON strings for weird symbols. So `{%json.data.fields.year}` would work, or `{%json.data.fields."my.horrible?bibtex.field.key"}`. This would line up with the proposed namespacing for keys / identifiers like `%canonical.provider` and `%canonical.sub_id`.

PR #329, opened 2026-07-17
Uniformize naming, clean up `autobib info`, and create terminology glossary

alexrutar:
This is a place to uniformize crate-wide naming, and also to create a terminology glossary.

So far, I have updated `autobib info` to by default output JSON. I also wrote JSON schemas for all of the JSON which Autobib outputs. We should make sure the crate internal naming lines up with what we are exposing in the schemas of course.


wupr:
I was about to open an issue on this, but since you're cleaning up `autobib info`, I'll put it here.

`autobib info` doesn't currently report the preferred key with `-r all`.

alexrutar:
I think it does in the version in this PR?

alexrutar:
Also, zbmath decided to change their API format again! This time, it's the 'null record' response. So I will fix that first.

alexrutar:
Previously, we used something like:

- `identifier`, or `id` short, is the blanket term (`Identifers` table, contains a bunch of identifiers!)
- `remote id` is a `provider:sub_id` (what about `local id`?)
- `canonical/reference id` is a special remote id

But, I also thought that `key` is reasonable - and then `bibtex key` is a key which is also valid bibtex. This lines up with the 'citation key' terminology common in bibtex world. 

I was using `preferred_keys` and `%key` because of the above. And then we would rename the `Identifiers` table to `Keys` table, for consistency.

I think an `identifier` should refer to a `provider:sub_id`. (Instead of "remote identifier", which is confusing because of `local:xxx`!) Then a remote identifier is anything not local, and a local identifier is itself. So we rename `RemoteId` to `Identifier`, inside the codebase.

alexrutar:
Other changes: 

- `RecordRow` used to refer to the row in the `Records` table. Now, it is just `Record`.
- `KeyedRecord` is a `Record` along with the `key` which was used to look it up.
- An entry is a record with usual data (not deleted or void)
- entry data, is the data itself
- raw entry data is the byte format
- whenever something external replies, it is called a `Response`. So a `DatabaseResponse`, is what the database says about a key; and a `RemoteResponse` is what a provider says about a key. All of those enums now have `Response` somewhere inside the name.

wupr:
> I think it does in the version in this PR?

Not when I tried half an hour ago. It reported `null` when `-r preferred` gives the correct key

alexrutar:
I also think we should change `%full_id` in templates to `%canonical`. It would be great if it were namespaced: like `%canonical.provider` and `%canonical.sub_id`, instead of `%provider` and `%sub_id` and `%full_id`. Related to #328

alexrutar:
> > I think it does in the version in this PR?
> 
> Not when I tried half an hour ago. It reported `null` when `-r preferred` gives the correct key

Weird let me try it out, it works for me.

alexrutar:
> > I think it does in the version in this PR?
> 
> Not when I tried half an hour ago. It reported `null` when `-r preferred` gives the correct key

Ah! Now I remember. Probably the new logic is confusing: `-r preferred` falls back to the canonical, if no preferred identifier could be found. However, `-r all` reports the canonical separately, and reports `preferred` as null.

I thought `-r preferred` should always return something usable (so that scripts can use it, without it being inconvenient); but it would be good to somehow show that there actually was no matching preferred key.

alexrutar:
Basically, `-r preferred` should always return something because my motivation for `-r preferred` was to do things like `autobib find | xargs autobib info -r preferred` which should always print something

wupr:
Ah!! That explains it. Then I think the `-r all` version should also have a fallback so as not to be confusing? And maybe the fallback could prioritise aliases instead, assuming users prefers aliases in general. I should really read the relevant code first, but my idea is that Autobib has an internal way of ranking keys which the user can override with config.

wupr:
I agree with the proposed terminology for identifiers and keys. It replaces the "mathematical" notion that "local" is a trivial "remote"; I (and I hope other mathematicians) really liked it, but I also appreciate it is confusing.

alexrutar:
> Ah!! That explains it. Then I think the `-r all` version should also have a fallback so as not to be confusing? And maybe the fallback could prioritise aliases instead, assuming users prefers aliases in general. I should really read the relevant code first, but my idea is that Autobib has an internal way of ranking keys which the user can override with config.

Since the new rules allow arbitrary regexes instead of only providers, you can implement this by adding a pattern which matches only aliases, like `[^:]+`. I'll add this suggestion to the "preferred_keys" config docs, and also document the canonical ID fallback explicitly.

alexrutar:
> I agree with the proposed terminology for identifiers and keys. It replaces the "mathematical" notion that "local" is a trivial "remote"; I (and I hope other mathematicians) really liked it, but I also appreciate it is confusing.

Sounds good! Yeah I agree, definitely my original logic too, but certainly it is confusing.

wupr:
> Since the new rules allow arbitrary regexes instead of only providers, you can implement this by adding a pattern which matches only aliases, like `[^:]+`. I'll add this suggestion to the "preferred_keys" config docs, and also document the canonical ID fallback explicitly.

That's what I added to my config yesterday to avoid having `null` for `"preferred"`, but it still feels like a workaround.

Maybe we can change the naming to distinguish whether there is fallback. In JSON it could be `"user_preferred"`?

wupr:
Two points about conventions.

The two `info`-like commands I use the most are `brew info` and `dnf info`; our original output format is more similar to them. So I think JSON output should be selected with a flag `--json`, as originally proposed.

It's more common to use camelCase for JSON fields. I think this is recommended in the Google style guide

alexrutar:
> Two points about conventions.
> 
> The two `info`-like commands I use the most are `brew info` and `dnf info`; our original output format is more similar to them. So I think JSON output should be selected with a flag `--json`, as originally proposed.
> 
> It's more common to use camelCase for JSON fields. I think this is recommended in the Google style guide

Fair enough - also, a `--json` flag could also mean we can input subsets of the JSON info for the other report types.

Hmm, I don't really like camelCase, even if it is recommended by Google. It's not a web API, and everything else we have prefers snake_case (Rust, SQLite, etc.)

alexrutar:
> > Since the new rules allow arbitrary regexes instead of only providers, you can implement this by adding a pattern which matches only aliases, like `[^:]+`. I'll add this suggestion to the "preferred_keys" config docs, and also document the canonical ID fallback explicitly.
> 
> That's what I added to my config yesterday to avoid having `null` for `"preferred"`, but it still feels like a workaround.
> 
> Maybe we can change the naming to distinguish whether there is fallback. In JSON it could be `"user_preferred"`?

I changed to `user_preferred` in the JSON.

alexrutar:
On second thought, I don't like the naming `user_preferred` very much. Who else could prefer it, if not the user?

alexrutar:
> The two `info`-like commands I use the most are `brew info` and `dnf info`; our original output format is more similar to them. So I think JSON output should be selected with a flag `--json`, as originally proposed.

Implemented now

alexrutar:
We still have columns `key` and `record_id` in the Records table. Should we rename these to `rev` and `canonical`, to align with the new naming?

alexrutar:
@wupr If you want to use an LLM for something, I think this is a good candidate task: If you agree, we should rename the columns in the Records table. This means modifying the migration code to include the following extra ALTER TABLE sql commands:
```
tx.execute("ALTER TABLE Records RENAME COLUMN record_id TO canonical", ())?;
tx.execute("ALTER TABLE Records RENAME COLUMN parent_key TO parent_rev", ())?;
tx.execute("ALTER TABLE Records RENAME COLUMN key TO rev", ())?;
```
The problem is that the column names are hard-coded everywhere inside the code... so there are a few hundred places where these things have to be renamed and finding them all is incredibly tedious. However, an LLM could probably do it precisely.

The main places where this should be changed are: `row.get(...)`, `row.get_unwrap(...)`, and the column names inside a SQL string including the Records table. But maybe there are others too. (But don't edit the migration code!)

Some indices in `src/db/schema/create_indices.sql` should be updated as well. (Existing indices will auto-rename, if I recall correctly.)

I wonder if we should do some compile-time concatenation using static variables:
```
TABLE_KEYS = "Keys"
TABLE_RECORDS = "Records"
TABLE_NULL_RECORDS = "NullRecords"
COL_RECORDS_REV = "rev"
COL_RECORDS_CANONICAL = "canonical"
... etc.
```
I don't know if this is worth it though, since it makes the SQL hard to read. Maybe we should just keep it as is.

Also as we previously discussed, it would be great to use an LLM to also to unify variable names. Even though many things were changed, I am sure there are some things that are missing.

alexrutar:
I used a bunch of regexes to do the renaming

wupr:
> On second thought, I don't like the naming `user_preferred` very much. Who else could prefer it, if not the user?

Autobib could. Autobib prefers canonical identifiers to all other identifiers by providing a fallback. If entirely neutral, it would just return a null result. In fact, when I read "preferred", I still first think of some ranking of the keys built into Autobib, before my brain corrects itself and realises that it's the user's preference. Perhaps this instinct comes from working with media streaming manifests, where the list of codecs might be ordered by the source to guide the player towards the most efficient option. The user could then of course override the order by expressing their codec preferences.

alexrutar:
> > On second thought, I don't like the naming `user_preferred` very much. Who else could prefer it, if not the user?
> 
> Autobib could. Autobib prefers canonical identifiers to all other identifiers by providing a fallback. If entirely neutral, it would just return a null result. In fact, when I read "preferred", I still first think of some ranking of the keys built into Autobib, before my brain corrects itself and realises that it's the user's preference. Perhaps this instinct comes from working with media streaming manifests, where the list of codecs might be ordered by the source to guide the player towards the most efficient option. The user could then of course override the order by expressing their codec preferences.

Ok, I am convinced! I have changed to `user_preferred`.

alexrutar:
I am going to merge this for now. I won't cut the new release yet, until then we can still make API changes.

wupr:
Oh, um, I had some pending review comments; I don't think I can submit them now. Maybe I'll open a PR instead.

One of the things I was going to suggest was merging the commit "Simplify autobib info somewhat" with "Large amounts of renaming for consistency", since the former undoes exactly the non-renaming changes in the latter.

alexrutar:
> Oh, um, I had some pending review comments; I don't think I can submit them now. Maybe I'll open a PR instead.
> 
> One of the things I was going to suggest was merging the commit "Simplify autobib info somewhat" with "Large amounts of renaming for consistency", since the former undoes exactly the non-renaming changes in the latter.

Oh, I didn't realize that happened! Sorry about that, I thought it would just work fine. I will be more careful next time / not merge like that again.

wupr:
Don't worry. I realise I've been taking my time with this one (today is the 9th consecutive day at a big conference; a bit overwhelmed).

I didn't find any major issue. I asked an LLM to go through the large-diff renaming commits and split out anything that was not strictly renaming. I then went through the PR myself commit by commit since I don't quite trust an LLM's judgement. I just have the 5 commits from yesterday left now.

I'll open a PR once I'm done.

alexrutar:
> Don't worry. I realise I've been taking my time with this one (today is the 9th consecutive day at a big conference; a bit overwhelmed).

Very long conference! It does sound very tiring. Absolutely, take your time, no rush. I have lots of free time now as you can probably tell...

PR #333, opened 2026-07-28
Migrate to external `autobib-entry` crate

### LINKED ISSUE #315: Zero-copy binary data backend?
Like the JSONB proposal, we could also use a zero-copy binary data format. I think we could use https://docs.rs/rkyv/latest/rkyv/collections/btree_map/struct.ArchivedBTreeMap.html

Essentially, the serialized format is byte-for-byte equivalent to the deserialized format, so there is no overhead at parse-time. There are more benefits:

1. `rkyv` canonically supports borrowed variants, so it would make it very cheap to read off specific fields without loading the whole data structure into memory. This is super useful e.g. for templates, and it would also let us implement very cheap / fast / memory efficient filtering code ("list all records which contain a title").
3. We don't have to maintain our own data format.

Downsides:

1. The data format is no longer documented.
2. Memory issues, if the data is corrupted
3. Still need the `MutableRecordData` type since zero-copy formats cannot be mutated.
4. More dependencies / worse compile time.

Another option instead of just migrating the backend, we could also allow multiple backends (see #309 or also a normal JSON backend).

### LINKED ISSUE #332: Field value too large
The record `doi:10.1103/PhysRevLett.114.191803` has so many authors it doesn't fit into our binary data format (try `autobib get doi:10.1103/PhysRevLett.114.191803`).

I guess the new binary data format should permit longer strings. Relevant for #309 and #315 or we change our binary data format locally.

alexrutar:
Summary:

1. Migrated to the new [`autobib-entry`](https://github.com/autobib/autobib-entry) crate.
2. Updated data format to `autobib_entry::v1` data format.
3. Cleaned up all of the data retrieval code to use a new 'typed `SELECT`' interface
4. Improved `autobib get` to borrow from the database instead of loading the data, whenever possible.


alexrutar:
Or put changes in a separate PR, and I will merge into this one.

wupr:
I'll make a separate PR (or rather several PRs) I think. What should I do with things that have been moved to the other crate in this PR? (Not sure if there are any such conflicts)

alexrutar:
Maybe it's easiest if you just edit this PR branch instead? I've got nothing unpushed. If it's in the new crate can PR there.

alexrutar:
Or feel free to merge this (it's in good state). And then stack other PRs on top.

wupr:
OK. I'll first make my commits on top of the current main. If there are anything affecting the new crate, we can patch there and bump this PR; otherwise I'll merge this PR.

alexrutar:
Even though there are no logic changes some of the type signatures changed, e.g. RawRecordData is now an unsized slice type (repr(transparent) wrapper around a [u8]) so there are a bunch of Box<RawRecordData> everywhere and signatures had to be modified to accept &D where D: EntryData basically everywhere (to permit unsized when needed). That's why there are a few hundred line edits.

alexrutar:
@wupr Just a heads up, the data backend migration is live! If you test it locally make sure to backup your DB. (I tested it on my own and it's fine.)

alexrutar:
I renamed `RawEntryData` to `ArchivedEntryData` (that's also what I used in `autobib-entry`).

alexrutar:
I rearranged `db.rs`: I created a new `DatabaseAccess` trait (implemented for `RecordDatabase` and `Snapshot`), which allows for operations which do not mutate the database, and I moved all of the mutating operations into `Snapshot`.

alexrutar:
I've implemented new generics for "types which can read from a row in the Records table". The key abstractions are
```rust
/// A trait for types which can access the data in a row in the Records table of the records database.
pub trait AccessRecordsRow<'row> {
    type Spec: spec::RowSpec;

    fn access_row(row: &'row Row<'_>) -> Self;

    /// Check that a statement contains all of the columns required to read this type
    /// from a row produced by the statement.
    fn check_statement(stmt: &rusqlite::Statement<'_>) -> bool {
        Self::Spec::check_statement_cols(stmt)
    }
}
```
and
```rust
/// A trait for types which can be loaded from the data in a row in the Records table.
///
/// This trait is implemented for any [`AccessRecordsRow`] implementation which does not capture the
/// `'row` lifetime. The reason for this lifetime bound is that the `'row` lifetime is managed by
/// the SQLite connection and cannot be safely captured by any return type.
pub trait LoadRecordsRow: for<'row> AccessRecordsRow<'row> {
    fn retrieve_from_rev(tx: &Tx<'_>, rev: RevisionId) -> Result<Option<Self>, rusqlite::Error>
    where
        Self: Sized,
    {
        use rusqlite::OptionalExtension;
        Self::retrieve_from_row_id(tx, rev.0).optional()
    }

    /// Load from an ID, with the caller promising that the ID exists in the database. For a
    /// fallible version, use `try_from_id`.
    fn retrieve_from_row_id(tx: &Tx<'_>, rev: i64) -> Result<Self, rusqlite::Error>
    where
        Self: Sized,
    {
        tx.prepare_cached(Self::Spec::as_records_select())?
            .query_row([rev], |row| Ok(Self::access_row(row)))
    }
}
```
with a blanket implementation
```rust
impl<T: for<'a> AccessRecordsRow<'a>> LoadRecordsRow for T {}
```
which means `LoadRecordsRow` is "free" as long as `AccessRecordsRow` does not capture the `'row` lifetime.

In order to verify 'composability' at the type level I added a trait
```rust
pub trait RowSpec {
    /// The columns that this row spec requires in order to read.
    fn col_names() -> &'static [&'static str];

    /// A select statement to retrieve this row spec from the records table, queried by row id.
    fn as_records_select() -> &'static str;

    /// Validate that this row spec is compactible with a `SELECT` statement specified in the
    /// compiled statement.
    fn check_statement_cols(stmt: &rusqlite::Statement<'_>) -> bool {
        let it = Self::col_names();
        let mut col_names = stmt.column_names();
        col_names.sort_unstable();
        it.iter().all(|k| col_names.binary_search(k).is_ok())
    }
}
```
which is morally "which columns does this type need, in order to be loaded from a SELECT statement". The trick is that `RowSpec` implementations form a DAG using `From` implementations so 'generic' types like `Record` can be implemented in a very canonical way like
```rust
impl<'r, D> AccessRecordsRow<'r> for Record<D>
where
    D: AccessRecordsRow<'r>,
    D::Spec: Into<spec::DataVariant>,
{
    type Spec = spec::RecordVariant;

    fn access_row(row: &'r Row<'_>) -> Self {
        let data = D::access_row(row);
        let canonical = Identifier::access_row(row);
        let modified = DateTime::access_row(row);

        Self {
            data,
            modified,
            canonical,
        }
    }
}
```
This means a `Record<D>` can be read from a SELECT statement, as long as the data type `D` can be read from a SELECT statement containing only 'data' and 'variant'.

Also, the `RowSpec` allows debug-time checking of the columns passed to the spec, so we can now use `debug_assert!` to check for these issues in tests without waiting for SQLite to error out if things are incorrect.

alexrutar:
Unfortunately I ran into some limits of Rust's trait system: I could not find a way to write down trait bounds on a closure generic over an argument `A` which does not capture the lifetime in the type `A`. Basically one can write closures `FnMut(&str) -> T` and then call this closure on very short lifetimes, but one cannot do the same for `FnMut(A) -> T`.

I think the problem is `FnMut(&str) -> T` desugars to `for<'a> FnMut(&'a str) -> T` but there is no `for<all possible lifetimes of D> FnMut(D) -> T`.

alexrutar:
I'm finished with this PR now and happy to merge as long as you agree with the database format change.

alexrutar:
I just realized I've had it backwards all along: the statement should hold the RowSpec! The access traits are correct but we instead need
```rust
trait TypedSelect {
    type Accepted: RowSpec;

    const STATEMENT: &'static str;
}
```
or similar. Then the interface is: user selects statement and return value, and the `RowSpec` are compared to check compatibility and `AccessRecordsData` is called to do the actual data conversion. This prevents the infinite proliferation of "access_x_as_y" methods.


I've always been thinking of "types should produce statements" but this is backwards.

alexrutar:
Sorry for editing this PR after I requested a review, I had an awesome idea for how to correctly express 'this X data type can be read from a SELECT statement with Y columns'. The trick is as follows:

Each "column" has a corresponding trait (not quite, because of the 'variant', but close enough). For example there are traits `col::Modified`, `col::Canonical`, etc.

A `SelectStatement` is a typed wrapper around a statement like `SELECT ... FROM ...`. Each `SelectStatement` implements the column traits corresponding to the values. For example, `SELECT modified, canonical FROM Records` would implement `col::Modified + col::Canonical`.

Dual to the `SelectStatement` is the `AccessRecordRow<'row, Q>` trait. The type parameter `Q` refers to a statement. A type, say, `Identifier<&'row str>`, implements `impl<'row, Q: col::Canonical> AccessRecordRow<'row, Q>`, which should be read as "An `Identifier` can be loaded from a `SELECT` statement as long as that statement implements `col::Canonical`".

Now the main trick which made me excited is that we get 'subtype select' for free: since `SELECT modified, canonical FROM Records` implements `col::Modified + col::Canonical` we can read an `Identifier`, or also a `DateTime<Local>`. This was the thing I was trying very hard to implement to avoid the 'N x M' issue: we have `N` implementations of `AccessRecordRow` and `M` implementations of `SelectStatement` and one can combine in a type-safe way any of these statements with any of these data types (as long as the trait bounds are compatible).

This removed quite a bit of boilerplate, and moreover since the `AccessRecordRow` traits are composable, it also removed all of the repetition for "this is a wrapper type, to read it, read its subtypes and then wrap them together". Like for `Record`, etc. So now we don't have `SELECT` statements and `from_row...` statements all over the place; they are all inside the `db::select` module. There are still some SQL statements elsewhere (for updates, etc. since those aren't typed since it is way too much work to make them typed) but the `SELECT` case is the main one I think is important for the future.

alexrutar:
The subtype selection thing fixed my previous idea (I hid all of the comments) because it avoids the "we need `2^N` traits for `N` columns" which got way too tedious and error-prone.

alexrutar:
Cut into 3 PRs now

PR #334, opened 2026-07-29
Improve documentation on JSON output

wupr:
This PR improves `docs/json.md` and the JSON schemas in `docs/schema`.

### DIFF
--- a/docs/schema/data.schema.json
+++ b/docs/schema/data.schema.json
@@ -15,6 +15,7 @@
     "fields": {
       "title": "Entry fields",
       "description": "The fields in the bibliographic data",
+      "type": "object",
       "additionalProperties": false,
       "patternProperties": {
         "^[^\\x00-\\x20{}(),=\\#%\"]+$": {
--- a/docs/json.md
+++ b/docs/json.md
@@ -5,7 +5,7 @@ This page documents the corresponding schemas.
 
 - `autobib source --json`: [`$source`](#source-autobib-source-output)
 - The `%json` meta key in templates: [`$record_entry`](#record_entry-record-of-type-entry)
-- `autobib info -r all --json`: [`$info`](#info-autobib-info-output)
+- `autobib info --json`: [`$info`](#info-autobib-info-output) for the `all` report type
 
 ## Schemas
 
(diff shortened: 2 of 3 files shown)


docs/schema/data.schema.json
```diff
@@ -10,16 +10,22 @@
       "title": "Entry type",
       "description": "The type of bibliographic entry",
       "type": "string",
-      "pattern": "^[^\\x00-\\x20{}(),=\\#%\"]+$"
+      "pattern": "^[^\\x00-\\x20{}(),=\\#%\"]+$",
+      "maxLength": 255
```
alexrutar: The new binary format will not have these length restrictions (see #332)

alexrutar (commented):
Just glancing through, nothing detailed.

docs/schema/data.schema.json
alexrutar: The new restriction will be `2^32 - 1` bytes though SQLite be unhappy well before this in practice.

alexrutar:
Other changes look good to me, thanks!

docs/schema/data.schema.json
wupr: OK, I'll remove the `maxLength`s then.

wupr:
Thanks! Merging now

docs/schema/data.schema.json
```diff
@@ -23,4 +24,4 @@
       }
     }
   }
-}
+}
```
alexrutar, line 27: Uhh, did this remove a trailing newline?

wupr: Oh oops that's my auto-formatter I think. I can put it back in my next PR. Is it better to have a trailing newline?

wupr: To clarify, the VS Code's default auto-formatter only seems to do this with JSON files

alexrutar: Weird, well trailing newline is POSIX standard so I think it's best! I fixed it already.

wupr: I see. I'd better change the VS Code settings for Autobib

PR #341, opened 2026-08-04
Fix some inconsistencies

wupr:


wupr:
These are some tiny fixes.

I have some other fixes related to naming (some of which were spotted by an LLM), but there are a few small things I wanted to discuss first, so they'll wait for another PR.

src/db/state/record.rs
```diff
@@ -126,7 +126,7 @@ impl<D: AsRecordData, T: AsRef<str>> Serialize for Record<D, T> {
     where
         S: serde::Serializer,
     {
-        let field_count = if self.data.serializable() { 2 } else { 3 };
+        let field_count = if self.data.serializable() { 3 } else { 2 };
```
alexrutar, line 129: Yikes, good catch! I think in practice the field count is only used by certain deserialize backends (but not JSON) but it could have resulted in something weird nonetheless.

alexrutar (approved):
looks good to me!

PR #343, opened 2026-08-04
New data backend

### LINKED ISSUE #315: Zero-copy binary data backend?

### LINKED ISSUE #332: Field value too large

alexrutar:
This PR adds a new data backend using the `autobib-entry` crate, and migrates to the new `v1` format.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


docs/changelog/next.md
```diff
@@ -1,18 +1,20 @@
...
-Supported database versions: `<= 4`
+Supported database versions: `<= 5`
 
 Changes since `v0.6.1`.
 
 ## Breaking changes
 
 - `zbmath` identifiers are now stored internally without 0-padding to length 8
+- The binary data format of the `data` column in the Records table has changed from the `v0` format to a new `v1` format.
+  Visit the [new entry data repository](https://github.com/autobib/autobib-entry) for detail on the new format.
```
wupr:
```suggestion
  Visit the [new entry data repository](https://github.com/autobib/autobib-entry) for details on the new format.
```

src/db/validate.rs
```diff
@@ -44,7 +44,9 @@ pub enum DatabaseFault {
     /// There was an underlying SQLite integrity error.
     IntegrityError(String),
     /// A row in the `Records` table contains invalid binary data.
-    InvalidRecordData(i64, String, AccessError),
+    InvalidRecordDataFormat(i64, String, AccessError),
+    /// A row in the `Records` table contains invalid binary data.
+    InvalidRecordData(i64, String, DataError),
```
wupr: The doc strings for these two variants are identical

src/db/validate.rs
```diff
@@ -132,10 +134,14 @@ impl fmt::Display for DatabaseFault {
...
             Self::IntegrityError(err) => write!(f, "Database integrity error: {err}"),
-            Self::InvalidRecordData(row_id, name, err) => write!(
+            Self::InvalidRecordDataFormat(row_id, name, err) => write!(
                 f,
                 "Record row '{row_id}' with record id '{name}' has invalid binary data: {err}"
             ),
+            Self::InvalidRecordData(row_id, name, err) => write!(
+                f,
+                "Record row '{row_id}' with record id '{name}' has data not in the standard format: {err}"
+            ),
```
wupr: The error messages look like they need to be swapped

src/db/migrate.rs
```diff
@@ -196,6 +196,68 @@ ALTER TABLE Keys RENAME COLUMN record_key TO record_rev;
...
+                "\
+                CREATE VIEW ActiveRecords \
+                AS \
+                  SELECT canonical, modified, data as entry_data \
+                  FROM Records \
+                  WHERE \
+                    rev in (SELECT record_rev FROM Keys) \
+                    AND variant = 0",
+                [],
+            )?;
```
wupr: Should we check if the view exists in `src/db/validate.rs`?

wupr (approved):

docs/changelog/next.md
wupr: Maybe we should move the first two bullet points here to the bottom of the section ~~(or a new section called internal changes)~~. In most cases, they don't require any user action or even knowledge.

Ah I guess they should be kept in this section, since they are incompatible with previous releases

alexrutar: I've moved them to the bottom and grouped them together.

src/db/validate.rs
alexrutar: It's correct, but the enum variant naming is not good. I renamed the variables so it is more clear.

alexrutar:
Some of the validation changes got tangled with other commits in the future PRs, so I will put them in the 'cleanup' ~commit~ PR #347

src/db/migrate.rs
alexrutar: Yes, and the indices too, I think.

